PR #5, opened 2019-04-12
Print LayoutUnits as pixels (e.g. "10.5px")

cbiesinger:


atanassov:
LGTM :) Thanks for adding it!

extensions/target-specific/chromium/blink/blink-helpers/blink-helpers.js
```diff
@@ -345,20 +345,24 @@ Loader.OnLoad(function() {
         });
     }));
 
+    function formatLayoutUnit(layout_unit) {
+      return (layout_unit / 64) + "px";
```
sanketj: Ideally, we would have queried for "kFixedPointDenominator" or "kLayoutUnitFractionalBits" using DbgObject.constantValue so that we wouldn't have to hard code the number 64 here. However, those symbol names don't seem to be present in the pdbs, which likely means they have been optimized out during compilation. So this is fine, but could you add a comment clarifying that 64 signifies the fixed point denominator?

extensions/target-specific/chromium/blink/blink-helpers/blink-helpers.js
```diff
@@ -345,20 +345,24 @@ Loader.OnLoad(function() {
         });
     }));
 
+    function formatLayoutUnit(layout_unit) {
```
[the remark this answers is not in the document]
cbiesinger: Done

extensions/target-specific/chromium/blink/blink-helpers/blink-helpers.js
cbiesinger: Done

cbiesinger:
Done, please take another look

extensions/target-specific/chromium/blink/blink-helpers/blink-helpers.js
```diff
@@ -345,20 +345,26 @@ Loader.OnLoad(function() {
         });
     }));
 
+    function layoutValueToPx(layout_unit) {
```
sanketj: nit: layout_unit should be layout_value?

PR #8, opened 2019-04-16
Add various LayoutNG types to blink-helpers

cbiesinger:


extensions/target-specific/chromium/blink/blink-helpers/blink-helpers.js
```diff
@@ -374,6 +384,18 @@ Loader.OnLoad(function() {
         });
     }));
 
+    DbgObject.AddArrayField(Chromium.RendererProcessType("blink::NGPhysicalFragment"), "children_", Chromium.RendererProcessType("blink::NGLinkStorage"), UserEditableFunctions.Create((fragment) => {
```
sanketj: I would suggest adding an extended field on NGPhysicalFragment that will perform a cast and return the object as a NGPhysicalContainerFragment. Users will be able to expand this field in the UI to see members of NGPhysicalContainerFragment in addition to members of NGPhysicalFragment, and you can invoke that field (via .F) as part of the "children_" array field implementation.

The "shareable_element_data_" and "unique_element_data_" extended fields on "blink::ElementData" do something similar.

cbiesinger:
OK, I made that change, let me know if that looks good to you. Wasn't sure on the naming of the extended field.

sanketj:
> OK, I made that change, let me know if that looks good to you. Wasn't sure on the naming of the extended field.

Yup, looks good. I think that naming convention is fine for fields that are just casting.

PR #10, opened 2019-04-16
Add an extension to view the NG fragment tree

cbiesinger:


cbiesinger:
Note that this only works with the extended field added by pull request #8

extensions/target-specific/chromium/blink/fragment-tree/fragment-tree.js
```diff
@@ -0,0 +1,78 @@
...
+                var errorMessage = ErrorMessages.CreateErrorsList(error) +
+                    ErrorMessages.CreateErrorReasonsList(ErrorMessages.WrongDebuggee("the Chromium renderer process"), ErrorMessages.SymbolsUnavailable);
+                return Promise.reject(errorMessage);
+            });
+        },
+        DefaultTypes: [Chromium.RendererProcessType("blink::NGPhysicalFragment")]
+    };
+
+    NGPhysicalFragmentTree.Tree.addChildren(Chromium.RendererProcessType("blink::NGPhysicalFragment"), (fragment) => {
+        return fragment.array("children_");
```
sanketj: What's the reason for including NGLinkStorage as part of the tree view? If it isn't necessary, it might be cleaner to just see a tree of NGPhysicalFragments.

extensions/target-specific/chromium/blink/fragment-tree/fragment-tree.js
```diff
@@ -0,0 +1,78 @@
...
+//
+//    Copyright (c) Google LLC. All rights reserved.
+//
+//--------------------------------------------------------------
+
+"use strict";
+
+var NGPhysicalFragmentTree = undefined;
+Loader.OnLoad(function() {
+    function fragmentToConcreteType(fragment) {
```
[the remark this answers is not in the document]
cbiesinger: Hm, you mean make three extended fields, one for each subclass? And then here return the first one that's not null?
I was hoping there's a better way to declare an extended field that may have different types, but I guess not :/

extensions/target-specific/chromium/blink/fragment-tree/fragment-tree.js
cbiesinger: NGLinkStorage stores the offset, which I thought was useful to have; and with the addNameRenderer call this doesn't look too bad, I thought (see also the screenshot on Slack #layoutng)

extensions/target-specific/chromium/blink/fragment-tree/fragment-tree.js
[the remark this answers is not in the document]
sanketj: For types without a v-table, there's unfortunately no better solution. For types that have a v-table though, you could just use vcast. A [vtable cast] extended field is automatically added in the type explorer for types that are vcast-able, so you wouldn't need to add your own fields for these.

From what I can see, NGPhysicalFragment doesn't have a vtable, but NGPhysicalContainerFragment does? So maybe you can get a NGPhysicalContainerFragment using your extended field from the other PR, and then vcast down? Then again, I think I'm looking a different set of symbols from you, because I don't see the "buffer_" member on NGPhysicalContainerFragment.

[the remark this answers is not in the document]
cbiesinger: I made the container fragment non-virtual yesterday, because it made it harder to deal with (casts did not do the right thing). That was in https://crrev.com/c/1569517

I will update this pull request shortly

extensions/target-specific/chromium/blink/fragment-tree/fragment-tree.js
sanketj: Yeah, I saw that. I thought it wasn't immediately obviously looking at the screenshot that the intermediary pieces were NGLinkStorage. Perhaps we should include the NGLinkStorage type name alongside the offset? Another option could be to display the offset alongside the NGPhysicalFragment types that are children of NGLinkStorage (though I don't know if there's a way to know this, since NGPhysicalFragment doesn't point back to its containing NGLinkStorage)?

cbiesinger: Done. Yes, I don't know how to avoid showing the NGLinkStorage intermixed though I'd love to find a way to do so.

extensions/target-specific/chromium/blink/fragment-tree/fragment-tree.js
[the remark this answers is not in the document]
cbiesinger: Done

extensions/target-specific/chromium/blink/fragment-tree/fragment-tree.js
sanketj: Another way is to add a 'synthetic' child to the tree, in the same way we add "Shadow Tree" or "Content Fragment" labels in the DOMTree extension. You could either make the NGLinkStorage node synthetic, or add synthetic children for each offset under NGLinkStorage and append the child physical fragment to that.

It's really up to you though if you want to keep it as is or go the synthetic child way. Just giving you anther option...

Issue #11, opened 2019-04-17
AXTree view is broken on recent Chromium builds


KurtCattiSchmidt:
This is due to the Chromium changes that introduced AXTreeManager.

sanketj:
When you complete the PR, I think it will give you an option to close this issue. If not, you can close it manually after.

KurtCattiSchmidt:
Resolved with PR https://github.com/MicrosoftEdge/JsDbg/pull/12

PR #12, opened 2019-04-17
Fixing AXTree code to refer to new AXTreeManager instance

KurtCattiSchmidt:
This change updates the AXTree code to use either the new AXTreeManager or the older g_ax_tree_id_map for constructing the AXTree.

### DIFF
--- a/extensions/promise/promise.js
+++ b/extensions/promise/promise.js
@@ -623,14 +623,14 @@ Using '.thenAll' makes this a bit smoother:\
     }
 
     Promise._help_any = {
-        description: "Similar to Promise.all, except it resolves when a single promise provided has succeeded.",
+        description: "Creates a single promise that resolves when any one of the promises in the given iterable resolve.",
         returns: "The succeeded promise.",
         arguments: [
-            {name:"promiseArray", type:"array", description: "The array of promises to evaluate."}
+            {name:"promiseIterable", type:"iterable", description: "The iterable collection (e.g. array) of promises to evaluate."}
         ]
     }
-    Promise.any = function(promiseArray) {
-        return Promise.all(promiseArray.map(p => {
+    Promise.any = function(promiseIterable) {
+        return Promise.all(promiseIterable.map(p => {
             // Invert Promise.all's rejection logic by swapping the pass/fail logic.
             return p.then(
               value => Promise.reject(value),
--- a/extensions/promise/promise.js
+++ b/extensions/promise/promise.js
@@ -621,4 +621,26 @@ Using '.thenAll' makes this a bit smoother:\
         
         return promisedType;
     }
+
+    Promise._help_any = {
+        description: "Similar to Promise.all, except it resolves when a single promise provided has succeeded.",
+        returns: "The succeeded promise.",
+        arguments: [
+            {name:"promiseArray", type:"array", description: "The array of promises to evaluate."}
+        ]
+    }
+    Promise.any = function(promiseArray) {
+        return Promise.all(promiseArray.map(p => {
+            // Invert Promise.all's rejection logic by swapping the pass/fail logic.
+            return p.then(
+              value => Promise.reject(value),
+              error => Promise.resolve(error)
+            );
+          })).then(
+            // If '.all' resolved, there were no resolved promises in the array
+            errors => Promise.reject(errors),
+            // If '.all' rejected, the error value will be the completed promise
+            value => Promise.resolve(value)
+          );
+    }
 })
--- a/extensions/promise/promise.js
+++ b/extensions/promise/promise.js
@@ -623,16 +623,16 @@ Using '.thenAll' makes this a bit smoother:\
     }
 
     Promise._help_any = {
-        description: "Creates a single promise that resolves when any one of the promises in the given iterable resolve.",
+        description: "Creates a single promise that resolves when any one of the promises in the given iterable resolves and rejects if all promises reject.",
         returns: "The succeeded promise.",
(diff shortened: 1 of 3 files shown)


KurtCattiSchmidt:
I created an issue here - https://github.com/MicrosoftEdge/JsDbg/issues/11, but I'm not sure how to link it to this PR.

extensions/target-specific/chromium/ax-tree/ax-tree.js
```diff
@@ -11,18 +11,25 @@
...
+            .catch(() => {
+                return DbgObject.global(Chromium.BrowserProcessSyntheticModuleName, "g_ax_tree_id_map").F("Object").array("Pairs").f("second");
+            })
+            .then((resolvedManagers) => {
+                managers = resolvedManagers;
+                return managers;
+            });
         }
-        return managersPromise;
+        return managers;
```
petersalas: This is kind of weird: sometimes `getManagers` returns a promise, sometimes it doesn't.  I'm guessing you went down this path because adding the `.catch` broke the `PromisedDbgObject` sugar?  I think better options are to either explicitly create a PromisedDbgObject from the promise, or change the callers to expect a regular Promise (which will resolve to a DbgObject) instead of a PromisedDbgObject.

KurtCattiSchmidt: You're right, this was kind of weird. It was basically a workaround to maintain the existing caching behavior while still handling possible exceptions from DbgObject.global.

The change I just pushed adds a Promise.all() method that eliminates the weird behavior and makes it much cleaner.

extensions/promise/promise.js
```diff
@@ -621,4 +621,26 @@ Using '.thenAll' makes this a bit smoother:\
         
         return promisedType;
     }
+
+    Promise._help_any = {
+        description: "Similar to Promise.all, except it resolves when a single promise provided has succeeded.",
```
sanketj: I don't think it similar to Promise.all at all. :)

I think it would be better to say something along the lines of: "Creates a single promise that resolves when any one of the promises in the given iterable resolve."

extensions/promise/promise.js
```diff
@@ -621,4 +621,26 @@ Using '.thenAll' makes this a bit smoother:\
...
     }
+
+    Promise._help_any = {
+        description: "Similar to Promise.all, except it resolves when a single promise provided has succeeded.",
+        returns: "The succeeded promise.",
+        arguments: [
+            {name:"promiseArray", type:"array", description: "The array of promises to evaluate."}
+        ]
+    }
+    Promise.any = function(promiseArray) {
```
sanketj: I would call the parameter "promises" or "promiseIterable", because what's passed in doesn't have to be an array.

extensions/target-specific/chromium/ax-tree/ax-tree.js
```diff
@@ -20,7 +20,19 @@ Loader.OnLoad(function() {
 
     function getManagers() {
         if (managersPromise == null) {
-            managersPromise = DbgObject.global(Chromium.BrowserProcessSyntheticModuleName, "g_ax_tree_id_map").F("Object").array("Pairs").f("second");
+            // g_ax_tree_id_map was replaced by a specialized AXTreeManagerMap. Check for both in case an older
```
sanketj: nit: older "build or" ...

extensions/promise/promise.js
```diff
@@ -621,4 +621,26 @@ Using '.thenAll' makes this a bit smoother:\
...
+        return Promise.all(promiseArray.map(p => {
```
sanketj: Love this API! :)

petersalas:
Sorry -- didn't realize I had to "submit my review" for my comments to show up!

extensions/promise/promise.js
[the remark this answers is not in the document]
sanketj: This is not the same as Promise.race. Promise.race resolves on first resolve and rejects on first reject. Promise.any resolves on first resolve, but rejects only when all the promises are rejected. This API is not in-built into the JS language, but Bluebird JS does implement this (http://bluebirdjs.com/docs/api/promise.any.html).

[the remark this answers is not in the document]
sanketj: @KurtCattiSchmidt We should probably clarify the rejection logic as part of the description.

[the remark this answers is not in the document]
KurtCattiSchmidt: Good call, the previous description sounded exactly like Promise.race because it left out the rejection behavior. Fixed.

extensions/target-specific/chromium/ax-tree/ax-tree.js
```diff
@@ -20,7 +20,19 @@ Loader.OnLoad(function() {
...
+            // crash dump is inspected.
+            managersPromise = Promise.any([
+                DbgObject.global(Chromium.BrowserProcessSyntheticModuleName, "instance", "base::NoDestructor<ui::AXTreeManagerMap>")
```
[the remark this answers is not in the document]
KurtCattiSchmidt: Yeah this is a bit weird, because it's caching a promised value, which I don't think is super common. But I think going forward, the Promise.any approach is a pretty clean way of guessing where a value in memory is stored - you can just have an array of possible Promise-based locations instead of a bunch of try/catch blocks.

extensions/promise/promise.js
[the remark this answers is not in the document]
petersalas: Ah of course -- my mistake.

PR #15, opened 2019-04-19
Inspector for a compositor frame

sushraja-msft:
This change implements a visualizer for compositorframe. Break into CompositorFrameSinkImpl::SubmitCompositorFrameInternal 
and then pass the address of &frame in the extension to see the contents of the frame.

on retail builds break into CompositorFrameSinkStubDispatch::Accept and pass 
&p_frame as the address to the extension.

Also included are some visualizers for gfx::Rect,gfx::Size and gfx::Transform

### DIFF
--- a/extensions/target-specific/chromium/chromium-helpers/chromium-helpers.js
+++ b/extensions/target-specific/chromium/chromium-helpers/chromium-helpers.js
@@ -34,6 +34,7 @@ var Chromium = null;
 
             RendererProcessName: "renderer",
             BrowserProcessName: "browser",
+            GpuProcessName: "gpu",
 
             _help_SetTargetProcess: {
                 description: "Sets the name of Chromium process being debugged.",
@@ -45,6 +46,7 @@ var Chromium = null;
 
             RendererProcessSyntheticModuleName: "renderer-module",
             BrowserProcessSyntheticModuleName: "browser-module",
+            GpuProcessSyntheticModuleName: "gpu-module",
 
             _help_RendererProcessType: {
                 description: "Creates a DbgObjectType in the Chromium renderer process.",
@@ -62,6 +64,14 @@ var Chromium = null;
             },
             BrowserProcessType: (typeName) => DbgObjectType(Chromium.BrowserProcessSyntheticModuleName, typeName),
 
+            _help_GpuProcessType: {
+                description: "Creates a DbgObjectType in the Chromium gpu process.",
+                arguments: [
+                    {name:"typeName", type:"string", description: "The type name."}
+                ]
+            },
+            GpuProcessType: (typeName) => DbgObjectType(Chromium.GpuProcessSyntheticModuleName, typeName),
+
             _help_RendererProcessEquivalentModules: {
                 description: "List of modules used in the Chromium renderer process.",
             },
@@ -73,6 +83,11 @@ var Chromium = null;
                 description: "List of modules used in the Chromium browser process.",
             },
             BrowserProcessEquivalentModules: ["msedge", "accessibility", "chrome"],
+
+            _help_GpuProcessEquivalentModules: {
+                description: "List of modules used in the Chromium gpu process.",
+            },
+            GpuProcessEquivalentModules: ["msedge", "service", "chrome"],
 
             _help_MultiProcessModules: {
                 description: "List of modules used in multiple Chromium processes.",
@@ -88,6 +103,8 @@ var Chromium = null;
                         SyntheticModules.RemoveEquivalentModules(Chromium.BrowserProcessSyntheticModuleName, ...Chromium.MultiProcessModules);
                     } else if (currentTargetProcessName == Chromium.RendererProcessName) {
                         SyntheticModules.RemoveEquivalentModules(Chromium.RendererProcessSyntheticModuleName, ...Chromium.MultiProcessModules);
+                    } else if (currentTargetProcessName == Chromium.GpuProcessName) {
+                        SyntheticModules.RemoveEquivalentModules(Chromium.GpuProcessSyntheticModuleName, ...Chromium.MultiProcessModules);
                     } else {
                         console.assert(false);
                     }
@@ -100,6 +117,8 @@ var Chromium = null;
                         SyntheticModules.AddEquivalentModules(Chromium.BrowserProcessSyntheticModuleName, ...Chromium.MultiProcessModules);
                     } else if (currentTargetProcessName == Chromium.RendererProcessName) {
                         SyntheticModules.AddEquivalentModules(Chromium.RendererProcessSyntheticModuleName, ...Chromium.MultiProcessModules);
+                    } else if (currentTargetProcessName == Chromium.GpuProcessName) {
+                        SyntheticModules.AddEquivalentModules(Chromium.GpuProcessSyntheticModuleName, ...Chromium.MultiProcessModules);
                     } else {
                         console.assert(false);
                     }
@@ -109,6 +128,7 @@ var Chromium = null;
 
         SyntheticModules.RegisterSyntheticName(Chromium.RendererProcessSyntheticModuleName, ...Chromium.RendererProcessEquivalentModules);
         SyntheticModules.RegisterSyntheticName(Chromium.BrowserProcessSyntheticModuleName, ...Chromium.BrowserProcessEquivalentModules);
+        SyntheticModules.RegisterSyntheticName(Chromium.GpuProcessSyntheticModuleName, ...Chromium.GpuProcessEquivalentModules);
 
         Help.Register(Chromium);
     });
--- a/extensions/target-specific/chromium/chromium-helpers/chromium-helpers.js
+++ b/extensions/target-specific/chromium/chromium-helpers/chromium-helpers.js
@@ -87,7 +87,7 @@ var Chromium = null;
             _help_GpuProcessEquivalentModules: {
                 description: "List of modules used in the Chromium gpu process.",
             },
-            GpuProcessEquivalentModules: ["msedge", "service", "chrome"],
+            GpuProcessEquivalentModules: ["msedge_child", "service", "chrome_child", "viz_common"],
 
             _help_MultiProcessModules: {
                 description: "List of modules used in multiple Chromium processes.",
--- a/extensions/target-specific/chromium/compositorframe/compositorframe.js
+++ b/extensions/target-specific/chromium/compositorframe/compositorframe.js
@@ -63,7 +63,7 @@ Loader.OnLoad(function () {
                     size--;
                   }
                 }
-                return Promise.all(quads).then((resolvedQuads) => { return resolvedQuads; });
+                return Promise.all(quads);
               });
             });
           });
@@ -85,7 +85,7 @@ Loader.OnLoad(function () {
   );
 
   DbgObject.AddTypeDescription(
-    DbgObjectType("viz_common", "viz::DrawQuad::Resources"),
+    Chromium.GpuProcessType("viz::DrawQuad::Resources"),
     "Resources",
     true,
     UserEditableFunctions.Create((r) => Promise.all([r.f("count").val()])
--- a/extensions/target-specific/chromium/compositorframe/compositorframe.js
+++ b/extensions/target-specific/chromium/compositorframe/compositorframe.js
@@ -88,11 +88,6 @@ Loader.OnLoad(function () {
     Chromium.GpuProcessType("viz::DrawQuad::Resources"),
     "Resources",
     true,
-    UserEditableFunctions.Create((r) => Promise.all([r.f("count").val()])
-      .thenAll((count) => {
-        return r.f("ids").vals(count).then((rs) => {
-          return rs;
-        });
-      }))
+    UserEditableFunctions.Create((r) =>  r.f("ids").vals(r.f("count").val()))
   );
 });
--- a/extensions/target-specific/chromium/compositorframe/compositorframe.js
+++ b/extensions/target-specific/chromium/compositorframe/compositorframe.js
@@ -31,13 +31,13 @@ Loader.OnLoad(function () {
   });
 
   CompositorFrame.Tree.addChildren(Chromium.GpuProcessType("viz::RenderPass"), (parentLayer) => {
-    return Promise.all([parentLayer.f("quad_list").f("helper_").f("data_").F("Object").pointerValue()]).then(
+    return parentLayer.f("quad_list").f("helper_").f("data_").F("Object").as("ntdll!void**").then(
       (charAllocator) => {
-        var voidObj = DbgObject.create("ntdll!void**", charAllocator);
-        if (!voidObj.isNull()) {
+
+        if (!charAllocator.isNull()) {
           // voidObj is a std::vector<std::unique_ptr<InnerList>>
-          return voidObj.size().then((platformPointerSize) => {
-            return voidObj.vals(2).then((val) => {
+          return charAllocator.size().then((platformPointerSize) => {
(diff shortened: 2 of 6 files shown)


extensions/target-specific/chromium/chromium-helpers/chromium-helpers.js
```diff
@@ -74,6 +84,11 @@ var Chromium = null;
             },
             BrowserProcessEquivalentModules: ["msedge", "accessibility", "chrome"],
 
+            _help_GpuProcessEquivalentModules: {
+                description: "List of modules used in the Chromium gpu process.",
+            },
+            GpuProcessEquivalentModules: ["msedge", "service", "chrome"],
```
dlibby-: these should be msedge_child and chrome_child

(can you also verify this works on a non-component build?)

extensions/target-specific/chromium/gfx-type-helpers/gfx-type-helpers.js
```diff
@@ -54,4 +54,51 @@ Loader.OnLoad(function() {
         UserEditableFunctions.Create((scrollOffset) => Promise.all([scrollOffset.f("x_").val(), scrollOffset.f("y_").val()])
             .thenAll((first, second) => `{${first}, ${second}}`))
     );
+
+    DbgObject.AddTypeDescription(
+        DbgObjectType("viz_common", "gfx::Rect"),
```
dlibby-: I'd like to get Sanket's take on how we should rationalize the process model here - ideally these new types would be supported for all processes (I suspect there is usage of these types in the browser process as well).

In the short-term though, can we use the GpuProcessType wrapper so that this works on non-component builds as well?

extensions/target-specific/chromium/compositorframe/compositorframe.js
```diff
@@ -0,0 +1,102 @@
...
+    );
+  });
+
+  CompositorFrame.Renderer.addNameRenderer(DbgObjectType("viz_common", "viz::SolidColorDrawQuad"), (solidColorDrawQuad) => {
+    return solidColorDrawQuad.f("color").val().then((c) => {
+      var a = ((c & 0xFF000000) >> 24) & 0X00000000000000ff;
+      var r = (c & 0x00FF0000) >> 16;
+      var g = (c & 0x0000FF00) >> 8;
+      var b = (c & 0x000000FF);
+      var color = "rgba(" + r + "," + g + "," + b + "," + a + ")";
```
dlibby-: alpha for css color is between 0 and 1, right?

sushraja-msft: Ah, thanks for pointing that out. Fixed it but couldnt validate since solid color quads dont have opacity.

extensions/target-specific/chromium/compositorframe/compositorframe.js
```diff
@@ -0,0 +1,102 @@
...
+    InterpretAddress: function (address) {
+      var voidObj = DbgObject.create("ntdll!void", address);
+      if (!voidObj.isNull()) {
+        return voidObj.as(Chromium.GpuProcessType("viz::CompositorFrame"));
+      }
+      return DbgObject.NULL;
+    },
+    GetRoots: function () {
+      return DbgObject.locals("service", "viz::CompositorFrameSinkImpl::SubmitCompositorFrameInternal", "frame").then(
+        (frames) => {
```
sanketj: The .then clause isn't doing anything. You can just return the result of DbgObject.locals.

extensions/target-specific/chromium/compositorframe/compositorframe.js
```diff
@@ -0,0 +1,102 @@
...
+          return frames;
+        }
+      );
+    },
+    DefaultTypes: []
+  };
+
+  CompositorFrame.Tree.addChildren(Chromium.GpuProcessType("viz::CompositorFrame"), (parentLayer) => {
+    return parentLayer.f("render_pass_list").array("Elements").map(unique_ptr => { return unique_ptr.F("Object"); });
```
sanketj: You can make **unique_ptr => { return unique_ptr.F("Object"); }** even cleaner, like this: **unique_ptr => unique_ptr.F("Object")**

extensions/target-specific/chromium/compositorframe/compositorframe.js
```diff
@@ -0,0 +1,102 @@
...
+  });
+
+  CompositorFrame.Tree.addChildren(Chromium.GpuProcessType("viz::RenderPass"), (parentLayer) => {
+    return Promise.all([parentLayer.f("quad_list").f("helper_").f("data_").F("Object").pointerValue()]).then(
+      (charAllocator) => {
+        var voidObj = DbgObject.create("ntdll!void**", charAllocator);
+        if (!voidObj.isNull()) {
+          // voidObj is a std::vector<std::unique_ptr<InnerList>>
+          return voidObj.size().then((platformPointerSize) => {
```
sanketj: Where are you getting platformPointerSize from?

extensions/target-specific/chromium/compositorframe/compositorframe.js
```diff
@@ -0,0 +1,102 @@
...
+  };
+
+  CompositorFrame.Tree.addChildren(Chromium.GpuProcessType("viz::CompositorFrame"), (parentLayer) => {
+    return parentLayer.f("render_pass_list").array("Elements").map(unique_ptr => { return unique_ptr.F("Object"); });
+  });
+
+  CompositorFrame.Tree.addChildren(Chromium.GpuProcessType("viz::RenderPass"), (parentLayer) => {
+    return Promise.all([parentLayer.f("quad_list").f("helper_").f("data_").F("Object").pointerValue()]).then(
+      (charAllocator) => {
+        var voidObj = DbgObject.create("ntdll!void**", charAllocator);
```
sanketj: Is it not possible to cast the returned value from F("Object") to ntdll!void** using .as?

extensions/target-specific/chromium/compositorframe/compositorframe.js
```diff
@@ -0,0 +1,102 @@
...
+          return voidObj.size().then((platformPointerSize) => {
+            return voidObj.vals(2).then((val) => {
+              // Memory address from val[0] to val[1] contain pointers
+              var count = (val[1] - val[0]) / platformPointerSize;
+              var address = val[0];
+              var innerListContentsArrayPromises = [];
+              for (var i = 0; i < count; i++) {
+                var innerList = DbgObject.create("ntdll!void**", address);
+                // Inner list has 4 values of interest data,capacity,size,step.
+                innerListContentsArrayPromises.push(innerList.val().then((addrInsertList) => {
```
sanketj: Isn't innerList.val() just address?

extensions/target-specific/chromium/compositorframe/compositorframe.js
```diff
@@ -0,0 +1,102 @@
...
+                  var capacity = innerListContentsArray[i][1];
+                  var size = innerListContentsArray[i][2];
+                  var step = innerListContentsArray[i][3];
+                  while (size > 0) {
+                    quads.push(DbgObject.create("service!viz::DrawQuad", data).vcast());
+                    data += step;
+                    size--;
+                  }
+                }
+                return Promise.all(quads).then((resolvedQuads) => { return resolvedQuads; });
```
sanketj: The quads array doesn't seem to contain promises, so it should be fine to just return quads here.

extensions/target-specific/chromium/gfx-type-helpers/gfx-type-helpers.js
sanketj: Yes, we should definitely wrap all of these with Chromium.GpuProcessType so that both component and non-component builds are supported.

In terms of supporting types in multiple processes...even if the same type exists in mutliple process, they are both technically distinct types. So I think we need a declarative way to add the same extended field/type description/array field on distinct types. This way we can share code and have the same helpers appear across processes. Perhaps, AddExtendedField/AddTypeDescription/AddArrayField could take an array of input types?

If this is not an immediate blocker, we can address this separately. Feel free to file an issue and just add the GpuProcessType wrapper for now.

extensions/target-specific/chromium/compositorframe/compositorframe.js
sushraja-msft: size()
Gets the size of the DbgObject in bytes.

so I called size when DbgObject is a void** to get the platform pointer size.

extensions/target-specific/chromium/compositorframe/compositorframe.js
sushraja-msft: turns out locals dont work reliabily, removing this code for now.

extensions/target-specific/chromium/compositorframe/compositorframe.js
sushraja-msft: the quads array contain promises to vcast, they need to be returned after being resolved.

extensions/target-specific/chromium/compositorframe/compositorframe.js
sushraja-msft: I like the code the way it is here since to allows me to inspect charAllocator.

extensions/target-specific/chromium/compositorframe/compositorframe.js
sushraja-msft: yes it is the address of the innerlist struct - whose symbol is not emitted in the pdb since it is defined in the .cc file. I cast to a void* and read the memory to get around this.

extensions/target-specific/chromium/compositorframe/compositorframe.js
sushraja-msft: actually resolved this.

extensions/target-specific/chromium/compositorframe/compositorframe.js
sanketj: No, I meant you don't have to get innerList.val(). You can just do this:

innerListContentsArrayPromises.push(DbgObject.create("ntdll!void**", address).vals(4))

extensions/target-specific/chromium/compositorframe/compositorframe.js
sanketj: Missed the vcast - thanks! In that case, can't you just return Promise.all(quads)?

extensions/target-specific/chromium/gfx-type-helpers/gfx-type-helpers.js
sanketj: You should still replace DbgObjectType with Chromium.GpuProcessType for all of these fields, and validate that they work with non-component builds.

extensions/target-specific/chromium/compositorframe/compositorframe.js
sanketj: Ah missed that..sorry!

extensions/target-specific/chromium/compositorframe/compositorframe.js
sanketj: Could you please file an issue regarding DbgObject.locals not working reliably? And it would great if you can include your example.

extensions/target-specific/chromium/gfx-type-helpers/gfx-type-helpers.js
sushraja-msft: I think I have replaced DbgObjectType in revision 2 ? I missed one in   CompositorFrame.js updating that now.

extensions/target-specific/chromium/compositorframe/compositorframe.js
sushraja-msft: I tried this again now, results in 404.

extensions/target-specific/chromium/compositorframe/compositorframe.js
sushraja-msft: that works.

extensions/target-specific/chromium/compositorframe/compositorframe.js
sushraja-msft: done.

extensions/target-specific/chromium/gfx-type-helpers/gfx-type-helpers.js
sushraja-msft: non component builds build with symbol level != 2, in such a build I cannot set break points on the compositor frame sink. So at the moment I cannot test this, though I have updated the process names like Daniel suggested.

extensions/target-specific/chromium/compositorframe/compositorframe.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+  );
+
+  DbgObject.AddTypeDescription(
+    Chromium.GpuProcessType("viz::DrawQuad::Resources"),
+    "Resources",
+    true,
+    UserEditableFunctions.Create((r) => Promise.all([r.f("count").val()])
+      .thenAll((count) => {
+        return r.f("ids").vals(count).then((rs) => {
```
sanketj: Why do you need the .then here?

extensions/target-specific/chromium/gfx-type-helpers/gfx-type-helpers.js
```diff
@@ -54,4 +54,51 @@ Loader.OnLoad(function() {
...
+        UserEditableFunctions.Create((size) => Promise.all([size.f("width_").val(), size.f("height_").val()])
+            .thenAll((first, second) => `{${first}, ${second}}`))
+    );
+
+    DbgObject.AddTypeDescription(
+        Chromium.GpuProcessType("gfx::Transform"),
+        "Size",
+        true,
+        UserEditableFunctions.Create((transform) => Promise.all([transform.f("matrix_").f("fmat").vals(16)])
+            .thenAll((mat) => {
```
sanketj: Do you need .thenAll here or can you just use .then?

extensions/target-specific/chromium/gfx-type-helpers/gfx-type-helpers.js
dlibby-: release (dev/canary) of msedge have symbol_level=2. It may be the case that that particular symbol is inlined, but you should still be able to test

sushraja-msft: E:\edgees\src>gn gen out/NoComp --args="is_component_build = false is_debug = true use_jumbo_build = false enable_nacl = false is_win_fastlink = true use_goma = true symbol_level = 2 goma_dir =\"%GOMACLIENTDIR%\""
ERROR at //build/config/compiler/compiler.gni:263:3: Assertion failed.
  assert(symbol_level != 2 || current_toolchain != default_toolchain ||
  ^-----
Can't do non-component debug builds at symbol_level=2

dlibby-: Right, but we want these things to work with release builds as well.

Is there anything debug specific to your extension?

sushraja-msft: Let me try release builds again, I had trouble getting a pointer to a compositor frame in release builds before.

extensions/target-specific/chromium/compositorframe/compositorframe.js
petersalas: `r.f("ids").vals(r.f("count").val())` will also suffice (DbgObject methods are generally very permissive about accepting promises).

extensions/target-specific/chromium/compositorframe/compositorframe.js
petersalas: @sushraja-msft out of curiousity, if the type wasn't emitted in the PDB, what type did it get without the cast?

PR #19, opened 2019-04-22
[gdb] Use binascii.hexlify for Python 2 compat

cbiesinger:
There's no hex() method in Python 2.

### DIFF
--- a/server/JsDbg.Gdb/JsDbg.py
+++ b/server/JsDbg.Gdb/JsDbg.py
@@ -2,6 +2,7 @@ import gdb
 import sys
 import subprocess
 import threading
+import binascii
 
 jsdbg = None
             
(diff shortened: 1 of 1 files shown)


MSLaguana:
I'm curious, what version of gdb are you using? I have

```
(gdb) pi
>>> import sys
>>> sys.version
'3.6.5 (default, Apr  1 2018, 05:46:30) \n[GCC 7.3.0]'
```

cbiesinger:
I have:
$ gdb --version
GNU gdb (GDB) 8.2.1
>>> sys.version
'2.7.16 (default, Apr  6 2019, 01:42:57) \n[GCC 7.3.0]'

MSLaguana:
Weird, that's a newer gdb with an older python. I guess I won't try to understand how they relate to each other.

cbiesinger:
I think I compiled this gdb myself (long story) but I don't know how it picked the python to use.

server/JsDbg.Gdb/JsDbg.py
```diff
@@ -248,7 +249,8 @@ def LookupSymbolName(pointer):
 def ReadMemoryBytes(pointer, size):
     inferior = gdb.selected_inferior()
     # Note: will throw an error if this includes unmapped/ unreadable memory
-    return inferior.read_memory(pointer, size).hex()
+    buf = inferior.read_memory(pointer, size)
+    return binascii.hexlify(bytearray(buf))
```
MSLaguana: Based on our discussions in slack, this might have unintended consequences since it adds some decorators to the string output.

cbiesinger: Please take another look - I fixed this with a version check

PR #21, opened 2019-04-22
[gdb] Fix off by one error

cbiesinger:
QueryDebuggerString returns the data without any prefix before
'{', so the data starts at 0, not 1.


sanketj:
@MSLaguana Was this just a bug or did QueryDebuggerString actually return a prefix for you? If it did return a prefix for you, could this difference be caused by different GDB versions?

MSLaguana:
I think this is a bug from when I did some refactoring. I updated most other indices to point to 0, but I missed these couple.

sanketj:
> I think this is a bug from when I did some refactoring. I updated most other indices to point to 0, but I missed these couple.

Ok, thanks.

PR #23, opened 2019-04-22
Use the htmlName in error messages

cbiesinger:
Otherwise, templates won't display correctly

### DIFF
--- a/extensions/dbgobject/arrays/arrays.js
+++ b/extensions/dbgobject/arrays/arrays.js
@@ -22,7 +22,7 @@ Loader.OnLoad(function() {
 
     ArrayField.prototype.ensureCompatibleResult = function(result, parentDbgObject) {
         if (!Array.isArray(result)) {
-            throw new Error("The \"" + this.name + "\" array on " + parentDbgObject.type.name() + " did not return an array.");
+            throw new Error("The \"" + this.name + "\" array on " + parentDbgObject.type.htmlName() + " did not return an array.");
         }
 
         var that = this;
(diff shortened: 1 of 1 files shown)


cbiesinger:
Is this the right place to do the escaping?

sanketj:
@cbiesinger Do you have an example of this issue?

cbiesinger:
Yeah, see e.g. this screenshot. I had a bug in an extended field, leading to an error message:
Screenshot from 2019-04-19 16-25-16

It should have been `std::__Cr::Map<Foo, Bar, Allocator<....> >`, but the `<Foo` part got interpreted as an HTML Tag.

sanketj:
> Yeah, see e.g. this screenshot. I had a bug in an extended field, leading to an error message:
> Screenshot from 2019-04-19 16-25-16
> 
> It should have been `std::__Cr::Map<Foo, Bar, Allocator<....> >`, but the `<Foo` part got interpreted as an HTML Tag.

Ah, ok. Thanks!

petersalas:
@sanketj I'm a bit late to this party, but shouldn't the downstream code that actually treats the error text as HTML escape it first (or use textContent etc.)

sanketj:
> @sanketj I'm a bit late to this party, but shouldn't the downstream code that actually treats the error text as HTML escape it first (or use textContent etc.)?

@petersalas I actually thought that was deliberate? Sorry, should've asked you first or dug a little deeper...

petersalas:
Nah, should've caught it earlier.  I'll look into it.

PR #24, opened 2019-04-23
[gdb] Also look up constants without the typename

cbiesinger:
So that the lookup works also for enums that are not
an enum class.

And use .cast instead of .reinterpret_cast.

Both of these seems to be required at least for
blink::Node::NodeFlags.


cbiesinger:
Could you merge this as well?

sanketj:
Was just waiting in case @MSLaguana wanted to have a look...

PR #25, opened 2019-04-23
Fix descriptions for single bit enums

sanketj:
Only single bit scalar values should get a true/false description. Single bit enums should work like all other enums.


sanketj:
@cbiesinger FYI...

PR #28, opened 2019-04-23
[gdb] Handle anonymous fields in GetAllFields and LookupField

cbiesinger:
This comes up here, for example:
https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/layout/layout_box.h?q=LayoutBox&sq=package:chromium&g=0&l=1812

Without this change we throw an exception, so we don't get any LayoutBox
fields. Longer-term, we need to find a way to handle anonymous unions.

### DIFF
--- a/server/JsDbg.Gdb/JsDbg.py
+++ b/server/JsDbg.Gdb/JsDbg.py
@@ -81,9 +81,10 @@ def DebuggerQuery(tag, command):
 
 
 class SFieldResult:
-    def __init__(self, field):
-        self.bitOffset = field.bitpos % 8 if hasattr(field, 'bitpos') else -1
-        self.offset = field.bitpos / 8 if hasattr(field, 'bitpos') else -1
+    # extra_bitoffset allows handling anonymous unions correctly
+    def __init__(self, field, extra_bitoffset=0):
+        self.bitOffset = (field.bitpos + extra_bitoffset) % 8 if hasattr(field, 'bitpos') else -1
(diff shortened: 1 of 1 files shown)


server/JsDbg.Gdb/JsDbg.py
```diff
@@ -151,7 +151,10 @@ def GetAllFields(module, type, includeBaseTypes):
             # Field is static
             continue
 
-        
+        if not field.name:
+            # Field is anonymous, we have no way to access it.
```
sanketj: I'm curious...how do we get this information in the debugger (GDB)?

cbiesinger: Looks like you access it as if it were a direct member of LayoutBox (like in C++):
(gdb) p $lo
$3 = (blink::LayoutBlockFlow *) 0x3587f6e22ab8
(gdb) p $lo->inline_box_wrapper_
$4 = (blink::InlineBox *) 0x0

So we could probably just consider them part of the parent, as long as we're careful with the bit offset...

cbiesinger: OK, please hold off on merging this, I really need to resolve this -- blink::Length is pretty useless without supporting anonymous unions

sanketj: Ok, will un-approve for now so it stays on my list.

cbiesinger: Done -- please take another look!

PR #29, opened 2019-04-23
[gdb] Support namespaces in LookupGlobalSymbol

cbiesinger:


extensions/target-specific/chromium/blink/blink-helpers/blink-helpers.js
```diff
@@ -551,7 +551,8 @@ Loader.OnLoad(function() {
             description: "Returns (a promise to) a collection of documents for all web frames in the current renderer process."
         },
         GetDocuments: () => {
-            return DbgObject.global(Chromium.RendererProcessSyntheticModuleName, "g_frame_map")
+            var module = Chromium.RendererProcessSyntheticModuleName;
+            return Promise.any([DbgObject.global(module, "g_frame_map"), DbgObject.global(module, "content::(anonymous namespace)::g_frame_map")])
```
sanketj: Is this an issue for all globals? Is there something we can change about DbgObject.global to avoid every caller having to do this?

cbiesinger: It's an issue for all globals that are defined in a namespace.

Hm, maybe we could have the JS code always pass the value with the namespace and have the server remove the namespace for windbg/visual studio?

Or yes, we could have DbgObject.global always do an automatic .any

Thoughts on those options?

sanketj: I think adding an optional namespace parameter to DbgObject.global is reasonable.

In general, the more we can do in JS the better. However, if there is a perf benefit to performing one symbolic lookup instead of two, then making a server side change may be worth considering.

Just to clarify, does the DIA global symbol lookup (what's used on Windows) not work if you look up a global symbol with its namespace?

cbiesinger: Yes, doesn't work with namespace in WinDbg:
```0:042> dt content_shell!g_frame_map
*** WARNING: Unable to verify checksum for content_shell.exe
DBGHELP: content_shell - private symbols & lines 
        c:\src\src\out\gn\content_shell.exe.pdb
   +0x000 private_instance_ : 0n140696262961808
   +0x008 private_buf_     : [16]  "???"
0:042> dt content_shell!content::(anonymous namespace)::g_frame_map
Symbol content_shell!content::(anonymous namespace)::g_frame_map not found.
0:042> dt content_shell!content::g_frame_map
Symbol content_shell!content::g_frame_map not found.
```

The advantage of doing it on the server is that it knows for sure whether the namespace is needed. But anyway, I'll add it to DbgObject.global for now.

cbiesinger: Please take another look -- now on DbgObject.global

sanketj: Now that I've thought about this some more, I think we should keep the namespace parameter on DbgObject.global, but do the rest of the handling server-side.

Instead of trying to look up both symbols and seeing which one works, we should pass the namespace parameter from DbgObject.global, through to LookupGlobalSymbol, then down to the JsDbg server into LookupGlobalSymbol on IDebugger. GdbDebugger and TypeCacheDebugger both implement IDebugger - TypeCacheDebugger can just ignore the namespace parameter and GdbDebugger can use it.

I did something similar to add the typename parameter to DbgObject.global recently (https://github.com/MicrosoftEdge/JsDbg/commit/2743129686132e507a4619a4971631a53c3b581c) - you can follow the same pattern.

I'd also recommend merging master into the GDB branch first to avoid merge conflicts. Let me know if it is already a nightmare - I can help out.

cbiesinger: OK, pull request for the merge is at https://github.com/MicrosoftEdge/JsDbg/pull/46

cbiesinger: https://github.com/MicrosoftEdge/JsDbg/pull/49 adds the API; once that's merged I will update this PR to implement it for GDB & use it in blink-helpers.

cbiesinger:
OK, updated. Please take another look.

I think it looks weird because it includes a merge from master to the gdb branch; I needed that for the namespace change.

server/JsDbg.Gdb/GdbDebugger.cs
```diff
@@ -227,8 +227,12 @@ class GdbDebugger : IDebugger {
...
         }
          
-        public async Task<SSymbolResult> LookupGlobalSymbol(string module, string symbol, string typename) {
-            string pythonResult = await this.QueryDebuggerPython(String.Format("LookupGlobalSymbol(\"{0}\",\"{1}\")", module, symbol));
+        public async Task<SSymbolResult> LookupGlobalSymbol(string module, string symbol, string typename, string nameSpace) {
+            string symbolName = "";
+            if (!String.IsNullOrEmpty(nameSpace))
+                symbolName = nameSpace + "::";
+            symbolName += symbol;
+            string pythonResult = await this.QueryDebuggerPython(String.Format("LookupGlobalSymbol(\"{0}\",\"{1}\")", module, symbolName));
```
sanketj: When a type name is passed in, do we need to do something more here to find the right symbol?

cbiesinger: I don't know how to get a list of all global symbols with a matching name in gdb... it usually disambiguates by filename, I think. https://sourceware.org/gdb/onlinedocs/gdb/Symbols.html#Symbols

Was there a specific symbol that prompted the addition of this parameter?

sanketj: Yes, the AXTree extension looks for chrome!instance, which has many occurrences, so the type name parameter is used to specify which one it wants. If you want to try that extension, you’ll need to enable accessibility settings in chrome://accessibility (there’s also a flag you can pass to Chromium during launch to enable these, but I don’t remember what it’s called), and you’ll have to attach to the browser process.

cbiesinger: Hmm, I see. I'll have to figure out something. Is it OK to address that separately though?

sanketj: If that's easier, sure. Are you going to work on that immediately? Otherwise, might be worth leaving a TODO so we don't forget about it.

cbiesinger: OK, added a TODO.

sanketj:
@cbiesinger Do you want me to merge master into jsdbg_gdb_support first? It might make this PR smaller and cleaner.

cbiesinger:
If you're fine with doing that, that would be great, thanks!

sanketj:
Sure, give me a few mins.

PR #31, opened 2019-04-23
[gdb] Support looking up enum values

cbiesinger:


server/JsDbg.Gdb/GdbDebugger.cs
```diff
@@ -120,11 +120,37 @@ class GdbDebugger : IDebugger {
...
+
+                string[] properties = fieldString.Split('#');
+                Debug.Assert(properties.Length == 2);
+                SConstantResult field = new SConstantResult();
+                field.ConstantName = properties[0];
+                field.Value = UInt64.Parse(properties[1]);
+                result.Add(field);
+
+                index = fieldEndIndex+1;
+                if (pythonResult[index] == ',') {
```
sanketj: Can you add a comment to document this case, similar to what you have on line 132?

cbiesinger: Done

sanketj: @cbiesinger I don't see the update here. Did you miss changing the PR?

server/JsDbg.Gdb/GdbDebugger.cs
```diff
@@ -120,11 +120,38 @@ class GdbDebugger : IDebugger {
...
-            // TODO:
-            //   -var-create - * ((type)constantValue)
-            //   may return a useful result; ^done,name="var1",numchild="0",value="ENUM_1",type="enum1",has_more="0"
-            return new List<SConstantResult>();
+            string pythonResult = await this.QueryDebuggerPython(String.Format("LookupConstants(\"{0}\", \"{1}\", {2})", module, type, constantValue));
+
+            List<SConstantResult> result = new List<SConstantResult>();
+
+            // TODO: Refactor this list parsing to share code with GetAllFields
+            // [{...}, {...}]
```
cbiesinger: I put it here, for consistency with where I put the other comment. Is that OK?

sanketj: Ah ok, missed that. Thanks!

PR #33, opened 2019-04-23
[gdb] Set Displacement to 0 if it is 2*pointer size

cbiesinger:
This accounts for the 2 words reserved for RTTI information. Because
the RTTI information is stored "before" the object itself, the vtable
pointer actually points to vtable + 2 pointers. See the link
in this commit for some more context.

DbgObject.symbol (and therefore DbgObject.vcast) requires displacement
to be 0 if you're logically at the beginning of the symbol.

### DIFF
--- a/server/JsDbg.Gdb/GdbDebugger.cs
+++ b/server/JsDbg.Gdb/GdbDebugger.cs
@@ -8,7 +8,7 @@ namespace JsDbg.Gdb {
     class GdbDebugger : IDebugger {
 
         public GdbDebugger() {
-            // TODO: Initialize from LookupTypeSize("void*")
+            // Assume 64-bit until we get a response from the debugger
             IsPointer64Bit = true;
         }
 
@@ -22,6 +22,8 @@ namespace JsDbg.Gdb {
         }
 
         public async Task Run() {
+            Task<uint> task = LookupTypeSize("", "void*");
+
             while(true) {
                 // Pump messages from python back to any waiting handlers
                 string response = await Console.In.ReadLineAsync();
@@ -29,6 +31,7 @@ namespace JsDbg.Gdb {
                     return;
                 }
                 this.OutputDataReceived?.Invoke(this, response);
+                IsPointer64Bit = task.Result == 8 ? true : false;
             }
         }
 
--- a/server/JsDbg.Gdb/GdbDebugger.cs
+++ b/server/JsDbg.Gdb/GdbDebugger.cs
@@ -8,6 +8,8 @@ namespace JsDbg.Gdb {
     class GdbDebugger : IDebugger {
 
         public GdbDebugger() {
+            // TODO: Initialize from LookupTypeSize("void*")
+            IsPointer64Bit = true;
         }
 
 
@@ -102,7 +104,6 @@ namespace JsDbg.Gdb {
             return result;
         }
 
-        // TODO: Determine whether the process is 64 bit or 32 bit        
         public bool IsPointer64Bit { 
             get {
                 return this.isPointer64Bit;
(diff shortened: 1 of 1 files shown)


server/JsDbg.Gdb/GdbDebugger.cs
```diff
@@ -8,6 +8,8 @@ namespace JsDbg.Gdb {
     class GdbDebugger : IDebugger {
 
         public GdbDebugger() {
+            // TODO: Initialize from LookupTypeSize("void*")
```
sanketj: Can we get this TODO fixed with (or before) this change? I think we should keep 32-bit and 64-bit working equally well at all times.

cbiesinger: Hm, it does look like nobody called IsPointer64Bit before. So the value was wrong but ignored... OK, I will fix that tomorrow.

server/JsDbg.Gdb/GdbDebugger.cs
```diff
@@ -320,7 +321,13 @@ class GdbDebugger : IDebugger {
 
             if (result.Name.StartsWith("vtable for ")) {
                 result.Name = result.Name.Substring("vtable for ".Length) + "::`vftable'";
-                // TODO: If displacement = 2 pointers, assume RTTI and set displacement to 0?
+                ulong pointer_size = IsPointer64Bit ? 8UL : 4UL;
+                if (result.Displacement == 2 * pointer_size) {
+                    // First two words of the vtable are reserved for RTTI.
+                    // http://refspecs.linuxbase.org/cxxabi-1.83.html#rtti-layout
+                    // TODO: Should we subtract 2*pointer_size from all vtable references?
+                    result.Displacement = 0;
```
MSLaguana: Will this have an adverse impact on other lookups? E.g. if a `bar` parameter is known to be at displacement `16`, is that consistently with/without the RTTI pointers included? If we have a 
```c++
struct TFoo {
  int a;
  int b;
  virtual void baz() {}
}

TFoo foo;
```

then if I `LookupSymbolName(&foo)` does that tell me that I'm at displacement `2*sizeof(void*)`, and is `&(foo->b)` at `2*sizeof(void*) + sizeof(int)`? If we lie and say that the former is at displacement 0, will that mess with trying to find `a` and `b` at some point?

I haven't looked at enough jsdbg code to know exactly how these results are used.

cbiesinger: It doesn't affect that at all. It's only about where the vtable pointer points to relative to the vtable symbol. For your example:

struct TFoo {
  void* implicit_vtable_pointer;
  int a;
  int b;
};
struct vtable_for_TFoo {
  void* some_rtti_stuff_1;
  void* some_rtti_stuff_2;
  void (*baz)();
};

implicit_vtable_pointer will not be at &vtable_for_TFoo, but will instead point to &vtable_for_TFoo.baz, and only that will be affected.

(The caller in question asks for LookupSymbolName TFoo.implicit_vtable_pointer, not &TFoo)

MSLaguana: I see, thanks for the clarification!

server/JsDbg.Gdb/GdbDebugger.cs
cbiesinger: OK, I updated this pull request to initialize IsPointer64Bit correctly. Please take another look!

server/JsDbg.Gdb/GdbDebugger.cs
```diff
@@ -20,6 +22,10 @@ class GdbDebugger : IDebugger {
         }
 
         public async Task Run() {
```
sanketj: The Run() function is called over and over right? If so, should we cache the result of LookupTypeSize?

sanketj: Or is it just called once and the loop below is doing all the work?

cbiesinger: Only once, see https://github.com/MicrosoftEdge/JsDbg/blob/jsdbg_gdb_support/server/JsDbg.Gdb/Program.cs#L34

sanketj: Ah yes. Of course...

PR #34, opened 2019-04-24
[gdb] Look for JsDbg under ~/JsDbg by default

cbiesinger:
This should be more likely to work for more people.

Probably still no long-term solution, though.

### DIFF
--- a/server/JsDbg.Gdb/JsDbg.py
+++ b/server/JsDbg.Gdb/JsDbg.py
@@ -3,6 +3,7 @@ import sys
 import subprocess
 import threading
 import binascii
+import os.path
 
 jsdbg = None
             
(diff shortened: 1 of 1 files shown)


server/JsDbg.Gdb/JsDbg.py
```diff
@@ -28,9 +29,10 @@ def __init__(self):
         self.showStderr = False
         self.verbose = True
         # TODO: assume that jsdbg is installed in "~/.jsdbg/" or some other known location?
```
sanketj: This is okay for now, but we should figure out a better solution or add documentation before we ship.

cbiesinger: Yes, I definitely agree

PR #37, opened 2019-04-24
[blink] Add an extended field for DataRef

cbiesinger:
To more easily go to the pointed object instead of requiring
two steps. Modeled after the std::unique_ptr extended field.

### DIFF
--- a/extensions/target-specific/chromium/blink/blink-helpers/blink-helpers.js
+++ b/extensions/target-specific/chromium/blink/blink-helpers/blink-helpers.js
@@ -10,6 +10,13 @@
 
 var BlinkHelpers = null;
 Loader.OnLoad(function() {
+    DbgObject.AddExtendedField(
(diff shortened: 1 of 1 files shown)


extensions/target-specific/chromium/blink/blink-helpers/blink-helpers.js
```diff
@@ -10,6 +10,13 @@
 
 var BlinkHelpers = null;
 Loader.OnLoad(function() {
+    DbgObject.AddExtendedField(
+        (type) => type.name().match(/^blink::DataRef<.*>$/) != null,
+        "Object",
```
sanketj: Should we call it "Object" or "Data"?

cbiesinger: Renamed to Data

PR #38, opened 2019-04-24
[gdb] Support debugger events

cbiesinger:
Also removes an unused variable.

### DIFF
--- a/server/JsDbg.Gdb/GdbDebugger.cs
+++ b/server/JsDbg.Gdb/GdbDebugger.cs
@@ -8,8 +8,7 @@ namespace JsDbg.Gdb {
     class GdbDebugger : IDebugger {
 
         public GdbDebugger() {
-            // Assume 64-bit until we get a response from the debugger
-            IsPointer64Bit = true;
+            OutputDataReceived += HandleGdbEvent;
         }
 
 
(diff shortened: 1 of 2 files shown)


server/JsDbg.Gdb/GdbDebugger.cs
```diff
@@ -36,6 +35,31 @@ class GdbDebugger : IDebugger {
...
+        void HandleGdbEvent(object sender, string ev) {
+            if (ev[0] != '%')
+                return;
+
+            bool oldState = debuggerBusy;
+
+            if (ev == "%cont")
+                debuggerBusy = true;
+            else if (ev == "%stop" || ev == "%exit")
+                debuggerBusy = false;
```
petersalas: It doesn't seem like we'll notify on the change from `Break` to `Detaching` -- is that intentional?

cbiesinger: I think you're right that we should, I'll fix that tomorrow.

BTW, can you confirm that I understood the semantics correctly? I.e. when you "continue" from a breakpoint we should send a waiting event?

petersalas: Yeah...`Waiting` is weird.  As you probably saw, in the WinDbg version it gets sent when we're polling for WinDbg to become available while trying to serve a request from the client.  On the client, when it gets the waiting event it throws up a full-page backdrop to say that progress is blocked by the fact that the debugger is running.  So sending it on every "continue" might cause more flickering than would be ideal? But it's not unreasonable.

cbiesinger: OK, changes made. Please take another look!

server/JsDbg.Gdb/GdbDebugger.cs
```diff
@@ -36,6 +35,27 @@ class GdbDebugger : IDebugger {
...
+
+            DebuggerChangeEventArgs.DebuggerStatus oldStatus = debuggerStatus;
+
+            if (ev == "%cont")
+                debuggerStatus = DebuggerChangeEventArgs.DebuggerStatus.Waiting;
+            else if (ev == "%stop")
+                debuggerStatus = DebuggerChangeEventArgs.DebuggerStatus.Break;
+            else if (ev == "%exit")
+                debuggerStatus = DebuggerChangeEventArgs.DebuggerStatus.Detaching;
+
```
sanketj: What about process and thread changes?

cbiesinger: You'll have to merge https://github.com/MicrosoftEdge/JsDbg/pull/46 before I can work on those :) The current jsdbg_gdb_support branch is too old for those events.

(But I'm happy to wait with merging this until after that pull request, and then I can add those too)

sanketj: Ah yes, of course. :)

Approved that merge. Can you just leave a TODO for those here?

cbiesinger: TODO added (in the python file)

PR #40, opened 2019-04-24
Support libc++ maps in stl-helpers

cbiesinger:
I based this on https://cs.chromium.org/chromium/src/third_party/libcxx-pretty-printers/printers.py
and it seems to work

### DIFF
--- a/extensions/target-specific/stl-helpers/stl-helpers.js
+++ b/extensions/target-specific/stl-helpers/stl-helpers.js
@@ -84,6 +84,53 @@ DbgObject.AddArrayField(
     }
 );
 
+DbgObject.AddArrayField(
+    (type) => {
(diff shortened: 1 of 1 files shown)


extensions/target-specific/stl-helpers/stl-helpers.js
```diff
@@ -106,7 +153,7 @@ DbgObject.AddArrayField(
 
 DbgObject.AddArrayField(
```
sanketj: Is it worth adding "Pairs" to std::map as well?

cbiesinger: Not sure I understand -- there's code around line 65 that adds Pairs to std::map

sanketj: Missed that - sorry! LGTM

PR #42, opened 2019-04-24
Enable support to continue debugging from the web frontend

sushraja-msft:
Adds feature to control the debugger from the front end. 
  - Mainly a productivity improvement as it avoids developer context switch. 

Adds  a setting in the treeinspector to continue the debugged process after the tree is updated.

  - This allows for a live tree view scenario that is really useful for debugging UI or watching trees update in response to complex changes in the application being debugged. Currently this is used for inspecting compositor frames during scrolling.

Notes:

In visual studio the debugger tries to take focus when it breaks in, leave the visual studio window snapped visible on screen to avoid it being brought to front and distract while debugging.


sanketj:
I'm not sure combining "update on break" and "continue" makes sense. I set a bp on the blink::Node constructor and loaded the DOMTree extension, and with "Update on break & continue", the tree view updates continuously and I couldn't use JsDbg until I turned updating off. Can we figure out a way to keep the two separate, while still achieving the debugging benefit you're looking for?

Feel free to discuss the design offline.

extensions/jsdbg/toolbar/toolbar.js
```diff
@@ -221,6 +221,15 @@ Loader.OnLoad(function () {
 
         toolbar.appendChild(threadSelectorPane);
 
+        var continueExecution = document.createElement("button");
```
sanketj: We should add a "step" button as well. I think it will be similarly valuable.

sushraja-msft: Yeah that sounds good, but as a seperate change perhaps.

extensions/tree-inspector/tree-inspector.js
```diff
@@ -399,6 +413,23 @@ var TreeInspector = (function() {
...
+        this.refreshOnBreakAndContinueGroupElement.appendChild(createElement("input", null, {
+            name: "refreshOnBreakAndContinue",
+            id: id("RefreshOnBreakAndContinue"),
+            type: "checkbox",
+        }, {
+            "change": function (e) {
+                that.setRefreshOnBreakAndContinue(e.target.checked);
+            }
+        }));
+        this.refreshOnBreakAndContinueGroupElement.appendChild(createElement("label", "Continue execution", {
```
sanketj: Instead of "Continue execution". can we say "Continue after next break"? Because the debugger doesn't automatically continue after checking that box.

extensions/tree-inspector/tree-inspector.js
```diff
@@ -399,6 +413,23 @@ var TreeInspector = (function() {
         updateCheckboxControl.appendChild(createElement("label", "Update When Debugger Breaks", {
             "for": id("RefreshOnBreak")
         }));
+        this.refreshOnBreakAndContinueGroupElement = createElement("span", null, {
```
sanketj: UI nit: Can we add a little bit more spacing between "Update When Debugger Breaks" and the checkbox for "Continue Execution"? They seem a little too close to each other.

extensions/tree-inspector/tree-inspector.js
```diff
@@ -96,6 +104,12 @@ var TreeInspector = (function() {
             .then(function () {
                 if (emphasisNodePtr != null) {
                     that.emphasizeNode(emphasisNodePtr, that.treeRoot, that.treeReader);
+                }                
+            })
+            .then(function () {
+                if (that.refreshOnBreak && that.refreshOnBreakAndContinueEnabled)
```
sanketj: Do we need to check for "refreshOnBreak" here? Won't "refreshOnBreakAndContinueEnabled" only be true if "refreshOnBreak" is true?

sanketj:
Two things:

(1) Can you and @cbiesinger coordinate on how we can add GDB support for this?

(2) It would be great if you could also add "step" while you're here. :)
     Feel free to do it as a separate change if thats easier.

cbiesinger:
Once this lands on master, I'm happy to merge to the gdb branch & add support for it there.

(If adding step, shouldn't next also be added?)

sanketj:
> Once this lands on master, I'm happy to merge to the gdb branch & add support for it there.
> 
> (If adding step, shouldn't next also be added?)

Yes, I meant both step into and step over. I guess step/next are the GDB terms. :)

server/JsDbg.WinDbg/DebuggerEngine.cs
```diff
@@ -75,6 +75,14 @@ class DebuggerEngine : IDiaDebuggerEngine {
             return await this.runner.GetCurrentProcessThreads();
         }
 
+        public Task Continue()
+        {
+            System.Diagnostics.Debug.WriteLine("Executing command: g");
```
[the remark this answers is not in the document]
sushraja-msft: Ah okay will do, I tested this scenario in the VS version earlier and nothing bad happened. Will wrap just in case.

extensions/tree-inspector/tree-inspector.js
sushraja-msft: "Continue after next break " doesnt explain behavior at each break. Continue execution is what visual studio calls it and the check box doesnt continue execution when you are currently broken in over there as well.

extensions/tree-inspector/tree-inspector.js
sushraja-msft: I dont want UI to control app logic, the check isnt harmful. Ill let it be.

extensions/tree-inspector/tree-inspector.js
sanketj: @sushraja-msft I agree that "Continue execution" is what the VS debugger calls it, but as someone who doesn't use that debugger by default, it wasn't intuitive. I expected "Continue execution" do actually perform a continue from my current bp. Not going to block on this, but it'd be great if we can come up with a better name for this concept.

Issue #39, opened 2019-04-24
Non-ASCII string displayed incorrectly


cbiesinger:
Steps to reproduce:
- Load http://blog.fefe.de/ in Chrome or Content Shell with LayoutNG enabled (ie. run with the --enable-blink-features=LayoutNG commandline flag)
  (did not test Edge)
- Run JSDbg
- Look at the Layout Tree extension
- Enable the "text" field on LayoutNGText (use "include base types")
The second LayoutNGText is missing the non-ASCII characters, e.g. shows "Were schne Verschwrungslinks" instead of "Wer schöne Verschwörungslinks"

Sorry, don't have time to take some screenshots

PR #44, opened 2019-04-25
WTFHelpers: Handle both Latin-1 and UTF-16 strings

sanketj:
There are two separate issues being fixed with this changes.

First, the .isUnsigned API on DbgObjectType is written incorrectly. This change fixes it work with unsigned values and all char types.

Second, the Text type description for WTF::StringImpl used to unconditionally treat the underlying string as being comprised of 8-bit characters. However, the string may contain 8-bit or 16-bit characters based on the is_8bit_ flag. This is also being fixed herewith so that the type description works for both Latin-1 and UTF-16 strings.

### DIFF
--- a/extensions/dbgobject/type/dbgobject-type.js
+++ b/extensions/dbgobject/type/dbgobject-type.js
@@ -79,8 +79,8 @@ Loader.OnLoad(function () {
     }
 
     dbgObjectType.prototype.isUnsigned = function() {
-        // Treat "char" as unsigned.
-        return this.name().indexOf("unsigned ") == this.name() == "char";
+        // Treat "char"s as unsigned.
+        return this.name().startsWith("unsigned ") || this.name().includes("char");
     }
 
     dbgObjectType.prototype.fullName = function() {
--- a/extensions/target-specific/chromium/blink/wtf-helpers/wtf-helpers.js
+++ b/extensions/target-specific/chromium/blink/wtf-helpers/wtf-helpers.js
@@ -21,7 +21,13 @@ Loader.OnLoad(function() {
     }));
 
     DbgObject.AddTypeDescription(Chromium.RendererProcessType("WTF::StringImpl"), "Text", true, UserEditableFunctions.Create((wtfStringImpl) => {
-        return !wtfStringImpl.isNull() ? wtfStringImpl.idx(1).as("char", /*disregardSize*/true).string(wtfStringImpl.f("length_")) : "";
+        if (!wtfStringImpl.isNull()) {
+            return wtfStringImpl.f("is_8bit_").val()
+            .then((is8bit) => wtfStringImpl.idx(1).as(is8bit ? "char" : "char16_t", /*disregardSize*/true))
+            .then((firstChar) => firstChar.string(wtfStringImpl.f("length_")));
+        } else {
+            return "";
+        }
     }));
 
     DbgObject.AddArrayField(


sanketj:
This will fix issue #39 .

extensions/dbgobject/core/dbgobject.js
```diff
@@ -814,6 +814,10 @@ Loader.OnLoad(function() {
             return "???";
         }
 
+        function signedOrUnsignedVals(character, length) {
+            return (character.type.name() == "unsigned char") ? character.uvals(length) : character.vals(length);
```
petersalas: I think `vals` is supposed to do this automatically?

extensions/dbgobject/core/dbgobject.js
```diff
@@ -855,9 +859,9 @@ Loader.OnLoad(function() {
                 };
                 return getRestOfString(that, "", /*suggestedChunkSize*/64);
             } else {
-                return that.vals(length)
+                return signedOrUnsignedVals(that, length)
                 .then(function (chars) {
-                    return chars.map(function (c) { return String.fromCharCode(c); }).join("");
+                    return chars.reduce((str, c) => str + String.fromCharCode(c), "");
```
petersalas: Isn’t this n^2?

extensions/target-specific/chromium/blink/wtf-helpers/wtf-helpers.js
```diff
@@ -21,7 +21,7 @@ Loader.OnLoad(function() {
     }));
 
     DbgObject.AddTypeDescription(Chromium.RendererProcessType("WTF::StringImpl"), "Text", true, UserEditableFunctions.Create((wtfStringImpl) => {
-        return !wtfStringImpl.isNull() ? wtfStringImpl.idx(1).as("char", /*disregardSize*/true).string(wtfStringImpl.f("length_")) : "";
+        return !wtfStringImpl.isNull() ? wtfStringImpl.idx(1).as("unsigned char", /*disregardSize*/true).string(wtfStringImpl.f("length_")) : "";
```
petersalas: I guess another option would be to change `string` to do use `uvals` unconditionally to make it more forgiving? Especially if `String.fromCharCode` assumes a certain size when given negative numbers.

petersalas: Sorry for all the questions — does `WTF::StringImpl` use UTF-8? Won’t this still be broken?

extensions/dbgobject/core/dbgobject.js
sanketj: Isn't .join('') also n^2? I think it depends on how these things are implemented under the hood and what optimizations can be applied. (I can imagine .join being written to not be n^2, but perhaps so can .reduce.) I ran this perf comparison of map().join() vs. reduce(): http://jsbench.github.io/#a80c1e0e7282af6742adbbea4dfdf50a. I found that it was only when I increased strings sizes to ~500000 characters that map().join() was faster, otherwise reduce() performed better. So maybe .join's optimization are only helpful when we have VERY large strings.

To be clear though, I wasn't making this change for perf reasons. :)
Reduce seems cleaner to me, though I don't have a super strong opinion about it...

extensions/dbgobject/core/dbgobject.js
sanketj: Ah yes, it should. Turns out there's a bug in isUnsigned - that's why it wasn't working.

extensions/target-specific/chromium/blink/wtf-helpers/wtf-helpers.js
sanketj: For the first question...open it doing to unconditionally, but just fixing isUnsigned solves the problem here.

For the second question...no, WTF::StringImpl uses Latin-1 or UTF-16 (https://chromium.googlesource.com/chromium/src/+/HEAD/third_party/blink/renderer/platform/wtf/text/README.md). The current implementation of this helper doesn't differentiate between the two, which might be a bug, but I haven't yet found a case with LCHARs or UCHARs where this doesn't work.

extensions/dbgobject/core/dbgobject.js
sanketj: Will pull this out of this change either way. If you think it is worth keeping, then I'll commit it separately.

cbiesinger: Yeah, this does not handle the Unicode case correctly for WTF::StringImpl (e.g. load http://www.google.com.hk with NG enabled and look at LayoutNGText::text_). But that was already broken, so could be fixed separately.

sanketj: Ah you're right...this doesn't work for UTF-16 sequences at all. Working on a fix - thanks!

extensions/target-specific/chromium/blink/wtf-helpers/wtf-helpers.js
sanketj: Should be fixed now.

extensions/dbgobject/core/dbgobject.js
sanketj: Both UTF-16 and Latin-1 should be handled correctly now.

Also String.fromCharCode can take a list of char codes, so we don't map().join() or reduce(). We can just spread the "chars" array.

extensions/dbgobject/type/dbgobject-type.js
```diff
@@ -80,7 +80,7 @@ Loader.OnLoad(function () {
 
     dbgObjectType.prototype.isUnsigned = function() {
         // Treat "char" as unsigned.
-        return this.name().indexOf("unsigned ") == this.name() == "char";
```
petersalas: Oops - looks like I botched this in refactoring a3420cf9 :(

extensions/target-specific/chromium/blink/wtf-helpers/wtf-helpers.js
petersalas: If UChar is 16-bits why is this code doing `.as("char")` at all?  Maybe it has the appearance of working because for ASCII characters it just ends up interpolating a bunch of null characters which don't get rendered?

I'd think that for either case if the `DbgObject` is correctly typed `.string(correctLengthInCountNotBytes)` would just work...

cbiesinger: It needs to do .as() because the string is stored "inline" in the memory right after the StringImpl. That memory needs to be interpreted as either 8-bit or 16-bit depending on whether StringImpl::is_8bit_ is true.

extensions/target-specific/chromium/blink/wtf-helpers/wtf-helpers.js
```diff
@@ -21,7 +21,13 @@ Loader.OnLoad(function() {
     }));
 
     DbgObject.AddTypeDescription(Chromium.RendererProcessType("WTF::StringImpl"), "Text", true, UserEditableFunctions.Create((wtfStringImpl) => {
-        return !wtfStringImpl.isNull() ? wtfStringImpl.idx(1).as("char", /*disregardSize*/true).string(wtfStringImpl.f("length_")) : "";
+        if (!wtfStringImpl.isNull()) {
+            return wtfStringImpl.f("is_8bit_").val()
+            .then((is8bit) => wtfStringImpl.idx(1).as(is8bit ? "char" : "wchar_t", /*disregardSize*/true))
```
cbiesinger: wchar_t is 4 byte on Linux -- would char16_t work?

extensions/target-specific/chromium/blink/wtf-helpers/wtf-helpers.js
sanketj: Right, LCHARs are 8-bit unsigned chars, whereas UCHARs are 16-bit.

extensions/dbgobject/core/dbgobject.js
petersalas: Why doesn't this (either what was here or what you have) work with UTF-16?

w.r.t. to reduce(...) vs. map(..).join(), I completely agree that reduce is cleaner -- my concern is just if there's a point at which the string concatentation explodes (1M+ characters isn't a ridiculous scenario.)  If you've convinced yourself that there's no such point, no objection.

extensions/target-specific/chromium/blink/wtf-helpers/wtf-helpers.js
sanketj: Good call! I wonder if we will need to update getTypeDescriptionFunctionIncludingBaseTypes in description.js to handle char16_t at some point.

extensions/dbgobject/core/dbgobject.js
```diff
@@ -856,9 +856,7 @@ Loader.OnLoad(function() {
                 return getRestOfString(that, "", /*suggestedChunkSize*/64);
             } else {
                 return that.vals(length)
-                .then(function (chars) {
-                    return chars.map(function (c) { return String.fromCharCode(c); }).join("");
-                });
+                .then(chars => String.fromCharCode(...chars));
```
petersalas: That would be even better, except it looks like "RangeError: Maximum call stack size exceeded" gets thrown above a certain point?

Like I said in the other thread, if you've convinced yourself there's effectively no point at which string concatenation explodes and want to change it, go for it.

sanketj: Yeah, just noticed that too. That's a no go! FWIW, even if we encounter 1M+ characters in a single string (which is pretty unlikely - even the HTML5 spec doesn't seem to have strings that are that big), .reduce shouldn't explode. I don't think this is a big deal though, so reverting to what it was!

extensions/dbgobject/core/dbgobject.js
sanketj: @petersalas It does work after fixing isUnsigned. Are you talking about something else?

PR #45, opened 2019-04-25
An extension to view Angle's D3D11 resource map

sushraja-msft:
This extension dumps the resource table within angle, this is useful to translate gl ids into actual directx textures. This extension only works on windows and when Angle uses a D3D11 backend.


extensions/target-specific/chromium/win-d3d11-angletextures/AngleTextures.js
```diff
@@ -0,0 +1,99 @@
...
+
+"use strict";
+
+var AngleTextures = undefined;
+Loader.OnLoad(function () {
+  AngleTextures = {
+    GetRoots: function () {
+      return DbgObject.global("libGLESv2", "gSingleThreadedContext").catch((error)=>{
+        var errorMessage = ErrorMessages.CreateErrorsList(error) +
+                    ErrorMessages.CreateErrorReasonsList("Extension works in windows only and when Angle uses D3D11 backend.", ErrorMessages.SymbolsUnavailable);
```
sanketj: Going forward, we should find a more declarative way of handling platform-specific extensions. Ideally, this extension wouldn't even show up on non-Windows platform. Perhaps we need to add something in extension.json. Can you file an issue for this?

extensions/target-specific/chromium/win-d3d11-angletextures/extension.json
```diff
@@ -0,0 +1,27 @@
+{
+    "name": "win-d3d11-angletextures",
```
sanketj: nit: please use the same naming format as the other extensions (WinD3D11AngleTextures)

extensions/target-specific/chromium/win-d3d11-angletextures/index.html
```diff
@@ -0,0 +1,76 @@
...
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <title>Inspect Compositor Frame</title>
+    <script type="text/javascript" src="/jsdbg/loader.js"></script>
+    <script type="text/javascript">
+        Loader.OnPageReady(() => {
```
sanketj: You should call Chromium.SetTargetProcess(Chromium.GpuProcessName) here.

extensions/target-specific/chromium/win-d3d11-angletextures/index.html
```diff
@@ -0,0 +1,76 @@
...
+                AngleTextures.ShowTextures(context) 
+            }, 
+            function (error) {
+                var rootsError = document.getElementById("rootsError");
+                rootsError.className = "error show";
+                rootsError.innerHTML = error;
+            });
+        });
+    </script>
+    <style>
```
sanketj: Consider moving these styles into a .css file.

extensions/target-specific/chromium/win-d3d11-angletextures/index.html
```diff
@@ -0,0 +1,76 @@
...
+    MIT License
+
+    Copyright (c) Microsoft Corporation. All rights reserved.
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <title>Inspect Compositor Frame</title>
```
sanketj: nit: consider using the extension name (or something similar) for the title

extensions/target-specific/chromium/win-d3d11-angletextures/index.html
```diff
@@ -0,0 +1,76 @@
...
+    <script type="text/javascript" src="/jsdbg/loader.js"></script>
+    <script type="text/javascript">
+        Loader.OnPageReady(() => {            
+            AngleTextures.GetRoots().then((context) => {   
+                document.getElementById("tables").className = "show";              
+                AngleTextures.ShowTextures(context) 
+            }, 
+            function (error) {
```
sanketj: nit: "(error) =>", to be consistent

extensions/target-specific/chromium/win-d3d11-angletextures/AngleTextures.js
sushraja-msft: I dont think this is tenable, for this extension you would need to know if Chrome is using SwiftShader or D3D11backend which you cannot tell unless you break in on the other hand if this is a general concept then If JsDBG intends to be untied from Chromium/Edge I don't think marking extensions as per platform would be important.

extensions/target-specific/chromium/win-d3d11-angletextures/extension.json
sushraja-msft: I think I am consistent with gfx-type-helpers, ax-tree, chromium-helpers. In cases where the words are merged they refer to class names in code ?

extensions/target-specific/chromium/win-d3d11-angletextures/index.html
sushraja-msft: done

extensions/target-specific/chromium/win-d3d11-angletextures/index.html
sushraja-msft: done

extensions/target-specific/chromium/win-d3d11-angletextures/index.html
sushraja-msft: done

extensions/target-specific/chromium/win-d3d11-angletextures/index.html
sushraja-msft: done

extensions/target-specific/chromium/win-d3d11-angletextures/AngleTextures.js
sanketj: Ideally, Windows-specific extensions shouldn't even be loaded/served on Mac/Linux (and vice versa). It's not just this one, we also have TEB and may have more in the future. They just add clutter on other platforms, so it's best to hide them from the user.

sanketj: We do something similar to hide the mshtml/edgehtml extensions. Seeing those in your view while debugging Chromium would be annoying. :)

extensions/target-specific/chromium/win-d3d11-angletextures/extension.json
sanketj: Both "gfx-type-helpers" and "chromium-helpers" are headless extensions (won't be directly visible to the user), so those are fine. For AXTree, though the file is called "ax-tree", the extension name is "AXTree" and that's what's shown on the JsDbg landing page. Please follow the same style.

PR #50, opened 2019-04-25
Fix bad merge - persistent store directory is gone

cbiesinger:
PersistentStore now always uses the temp directory


sanketj:
Sorry :(

PR #61, opened 2019-04-29
[blink-helpers] Add new names for NGPhysical{Offset,Size}

cbiesinger:
They were renamed in https://crrev.com/c/1582872

### DIFF
--- a/extensions/target-specific/chromium/blink/blink-helpers/blink-helpers.js
+++ b/extensions/target-specific/chromium/blink/blink-helpers/blink-helpers.js
@@ -480,9 +480,14 @@ Loader.OnLoad(function() {
         .thenAll((first, second) => `{${layoutValueToPx(first)}, ${layoutValueToPx(second)}}`);
     }));
 
-    DbgObject.AddTypeDescription(Chromium.RendererProcessType("blink::NGPhysicalSize"), "Size", true, UserEditableFunctions.Create((size) => {
-        return Promise.all([size.f("width").val(), size.f("height").val()])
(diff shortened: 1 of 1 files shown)


extensions/target-specific/chromium/blink/blink-helpers/blink-helpers.js
```diff
@@ -480,21 +480,33 @@ Loader.OnLoad(function() {
         .thenAll((first, second) => `{${layoutValueToPx(first)}, ${layoutValueToPx(second)}}`);
     }));
 
+    // Old name for PhysicalSize
     DbgObject.AddTypeDescription(Chromium.RendererProcessType("blink::NGPhysicalSize"), "Size", true, UserEditableFunctions.Create((size) => {
         return Promise.all([size.f("width").val(), size.f("height").val()])
         .thenAll((first, second) => `{${layoutValueToPx(first)}, ${layoutValueToPx(second)}}`);
     }));
 
+    DbgObject.AddTypeDescription(Chromium.RendererProcessType("blink::PhysicalSize"), "Size", true, UserEditableFunctions.Create((size) => {
```
sanketj: I think we can avoid writing the extended field twice by using a function that regex matches both blink::NGPhysicalSize and blink::PhysicalSize (like we do in stl-helpers).

cbiesinger: Done. I thought Chromium.RendererProcessType would be a problem but it wasn't.

PR #63, opened 2019-04-29
gfx-type-helpers: Use a function that does regex matching for gfx::Size to share the same extended field between processes

sanketj:


sanketj:
@dlibby- / @sushraja-msft - let me know if we should do this for any other gfx types

dlibby-:
I think it'd be good to change this for all the gfx:: types in the file currently.

sanketj:
> I think it'd be good to change this for all the gfx:: types in the file currently.

@dlibby- Done!

PR #64, opened 2019-04-29
[blink-helpers] Add some more LayoutNG types

cbiesinger:


cbiesinger:
No current plans to do so. The Logical/Physical ones were a special case because they were moved to a more general location.

PR #67, opened 2019-04-30
Move instructional videos out of JsDbg package

sanketj:
The videos are quite large in size and slow down the downloading of JsDbg. They have now been moved out of JsDbg itself, and into a YouTube channel. A new markdown file, VIDEOS.md, is being added herewith that points to each of the videos.


cbiesinger:
Thanks!

PR #70, opened 2019-04-30
Handle failed loads of .js files more gracefully

cbiesinger:
Currently, if a JS file fails to load we hang forever even if
the current page actually doesn't need it. It seems better to
also add an error handler to consider it finished even if it
fails.

This is prompted by me noticing that angletextures.js is not loading
with a 404 error, which I have not debugged yet.


cbiesinger:
The underlying problem will be fixed in https://github.com/MicrosoftEdge/JsDbg/pull/72

PR #72, opened 2019-04-30
[angletextures] Fix capitalization of angletextures.js

cbiesinger:
Otherwise, the JS file fails to load on Linux, which is bad even if
the extension doesn't work there anyway.


sushraja-msft:
Thanks Christian 👍

PR #73, opened 2019-04-30
[gdb] Fix most compiler warnings in GdbDebugger.cs

cbiesinger:


sanketj:
@cbiesinger Could you clarify which compiler you are using and what are the errors?

cbiesinger:
I'm using dotnet core (from https://dotnet.microsoft.com/download); for the code I'm changing here it gives these warnings:
GdbDebugger.cs(510,17): warning CS4014: Because this call is not awaited, execution of the current method continues before the call is completed. Consider applying the 'await' operator to the result of the call. [/usr/local/google/home/cbiesinger/JsDbg/server/JsDbg.Gdb/JsDbg.Gdb.csproj]

sanketj:
Ah okay, I see the same in VS now. Just wanted to make sure this was consistent across compilers.

PR #75, opened 2019-05-01
[gdb] Add debugger messages so the user can see progress

cbiesinger:
Similar to what the Windows implementation outputs.

I also turned off stderr output by default because otherwise
the GDB window gets very crowded with the message output.


sanketj:
Nice! :)

PR #76, opened 2019-05-01
[gdb] Various logging output improvements

cbiesinger:
- Don't print debugger messages to stderr on Linux, but keep other stderr
  output. Rationale: debugger messages get too verbose because they get
  intermixed with your gdb session, but other stderr output is useful,
  in particular the server URL gets printed to stderr.
- If the server crashes, show a message to the user and reset the jsdbg
  global, so that typing jsdbg again will do the right thing.

### DIFF
--- a/server/JsDbg.Gdb/Program.cs
+++ b/server/JsDbg.Gdb/Program.cs
@@ -22,6 +22,8 @@ namespace JsDbg.Gdb
             PersistentStore persistentStore = new PersistentStore();
 
             using (WebServer webServer = new WebServer(debugger, persistentStore, extensionsDirectory)) {
+                // Turn off printing of debugger messages, because they fill up the gdb console too quick.
+                webServer.PrintDebuggerMessages = false;
                 webServer.LoadExtension("default");
 
                 try {
--- a/server/JsDbg.Core/WebServer.cs
+++ b/server/JsDbg.Core/WebServer.cs
@@ -119,7 +119,8 @@ namespace JsDbg.Core {
             this.debugger = debugger;
             this.debugger.DebuggerChange += (sender, e) => { this.NotifyClientsOfDebuggerChange(e.Status); };
             this.debugger.DebuggerMessage += (sender, message) => {
-                Console.Error.WriteLine(message);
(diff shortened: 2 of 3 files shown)


server/JsDbg.Gdb/Program.cs
```diff
@@ -22,6 +22,8 @@ class Program
             PersistentStore persistentStore = new PersistentStore();
 
             using (WebServer webServer = new WebServer(debugger, persistentStore, extensionsDirectory)) {
+                // Turn off printing of debugger messages, because they fill up the gdb console too quick.
```
sanketj: Is the expectation that the user has to make a code change to get debugger messages? Is there a flag or something that we can pass to JsDbg on launch to make this easier?

cbiesinger: The Web UI shows them (in the progress bar at the bottom)

sanketj: That's true, but I think the console messages will still be useful when debugging the server (especially with WSL). I think it's fine if the messages are off by default, but it would be good to have a convenient way to turn them on.

cbiesinger: (what's WSL?)
For debugging the server, you can turn on the "verbose" option in the python script or through gdb (python JsDbg.jsdbg.verbose=True), which has similar information to the debugger messages but more detail.

sanketj: Windows Subsystem for Linux. I guess setting verbose=True is convenient enough. As long as it captures all the same information, it's fine.

In that case though, is there any value in having the PrintDebuggerMessages member? It looks like it will never be set to true apart from dispose.

cbiesinger: Well I did that so that on Windows those messages will still be printed. I'm not sure I understand your Dispose reference -- it defaults to true and for Gdb it gets set to false.

sanketj: Oh I see...the PrintDebuggerMessages member is in JsDbg.Core. I misread and thought it was GDB specific. My mistake...

PR #84, opened 2019-05-03
[gdb] Fix bugs in GetCallStack and GetSymbolsInStackFrame

### LINKED ISSUE #77: [gdb] DbgObject.locals does not work
Returns an empty array.

Looking at the verbose output, it looks like it works by calling GetCallStack followed by LookupSymbolName, and the latter does not currently work for stack variables

cbiesinger:
Fixes #77

### DIFF
--- a/server/JsDbg.Gdb/GdbDebugger.cs
+++ b/server/JsDbg.Gdb/GdbDebugger.cs
@@ -355,6 +355,7 @@ namespace JsDbg.Gdb {
             while(pythonResult[index] != ']') {
                 // '{%s#%d#%s}' % (self.name, self.symbolResult.pointer, s.symbolResult.type)
                 Debug.Assert(pythonResult[index] == '{');
+                ++index;
 
                 int symEndIndex = pythonResult.IndexOf('}', index);
                 string symString = pythonResult.Substring(index, symEndIndex - index);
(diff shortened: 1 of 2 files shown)


cbiesinger:
I've repurposed this pull request to fix all the issues in GetCallStack and GetSymbolsInStackFrame; they work for me now when I call JsDbg.GetCallStack / JsDbg.LookupLocalsInStackFrame

server/JsDbg.Gdb/GdbDebugger.cs
```diff
@@ -304,11 +304,11 @@ class GdbDebugger : IDebugger {
             // NOTE that quotes can also appear in filenames, but they will be \ escaped
 
             string pythonResult = await this.QueryDebuggerPython(String.Format("GetCallStack({0})", frameCount));
-            
 
             List<SStackFrame> result = new List<SStackFrame>();
 
-            int index = 1; // Skip initial '"' character
```
sanketj: Was this just a bug before or has something changed such that now we don't see the '"' character?

cbiesinger: I'm fairly sure that this is basically the same issue as https://github.com/MicrosoftEdge/JsDbg/pull/21

Maybe @MSLaguana could comment better

MSLaguana: Yeah I think it is the same; probably from when I switched from parsing the gdb mi format to doing everything via python.

PR #85, opened 2019-05-03
[gdb] Fix the last remaining compiler warning

cbiesinger:
GdbDebugger.cs(534,34): warning CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread. [.../JsDbg/server/JsDbg.Gdb/JsDbg.Gdb.csproj]

### DIFF
--- a/server/JsDbg.Gdb/GdbDebugger.cs
+++ b/server/JsDbg.Gdb/GdbDebugger.cs
@@ -531,7 +531,9 @@ namespace JsDbg.Gdb {
             string response = await this.QueryDebuggerPython("GetCurrentProcessThreads()");
             return ParsePythonArrayToIntegers(response).ToArray();
         }
-        public async Task<ulong> TebAddress() { return 0; }
+
+        public Task<ulong> TebAddress() { return Task.FromResult(0UL); }
+
         public uint TargetProcess {
             get {
                 return this.targetProcess;


server/JsDbg.Gdb/GdbDebugger.cs
```diff
@@ -531,7 +531,9 @@ class GdbDebugger : IDebugger {
             string response = await this.QueryDebuggerPython("GetCurrentProcessThreads()");
             return ParsePythonArrayToIntegers(response).ToArray();
         }
-        public async Task<ulong> TebAddress() { return 0; }
+
+        public Task<ulong> TebAddress() { return Task.FromResult(0UL); }
```
sanketj: Not related to this change...TEB is a Windows-only concept, so it doesn't make sense to show this extension on other platforms. I think we should have a declarative way to write platform-specific extensions. Perhaps we can add a "targetPlatforms" field to the extension manifest? I'll file an issue to track this.

cbiesinger: I believe the TEB extension is already limited to the "ntdll" module

sanketj: That's true, forgot about that...
So it isn't a problem in this case. But I think this is applicable to the WinD3D11AngleTextures extension, and other extensions we may right for platform-specific technologies in the future. The TSF extension (not checked-in, but there's a branch for it) is another example. Doesn't make sense to have these on Linux/Mac.

cbiesinger: Yeah, I agree.

sanketj: Filed issue #86 .

PR #93, opened 2019-05-07
[fragment tree] Gracefully handle a null LayoutView or null first child

cbiesinger:


sanketj:
For this extension, could you also add an error message mentioning that LayoutNG needs to be enabled? Feel free to do this in a separate change if that makes more sense.

sanketj:
Also, should we add null LayoutView filtering to the LayoutObjectTree extension as well?

cbiesinger:
Thanks. I'll do it as a separate change.

I think for LayoutObjectTree it is less important because we don't access any members of it, but I guess we might show a null root without filtering?

sanketj:
> I think for LayoutObjectTree it is less important because we don't access any members of it, but I guess we might show a null root without filtering?

Looking at this some more, it seems that documents only get added to _g_frame_map_ after they get an associated LayoutView. So I'm not sure there's a case where we'll get back a document from BlinkHelpers.GetDocuments() that won't have an associated LayoutView. If you know of such a case, I'd love to hear what it is.

cbiesinger:
Pull request #94 adds the error message when LayoutNG is disabled

As for null LayoutView -- Looking at Document::Initialize, it does seem true that all of them have a LayoutView. This seems to be called shortly *after* the document gets added to g_frame_map, but both happen from WebLocalFrameImpl::CreateMainFrame --  MakeGarbageCollected<WebLocalFrameImpl> triggers the g_frame_map addition, frame->InitializeCoreFrame triggers the LayoutView constructor (through a deep call stack).

So, I think you're right that the LayoutView nullcheck is not necessary and I will remove it in a separate pull request

PR #96, opened 2019-05-07
[gdb] Add build and debugging instructions

cbiesinger:
Mostly for when we merge this branch to master.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -15,6 +15,8 @@ The JsDbg source can be obtained from the [JsDbg repository on GitHub](https://a
 
 ### Building JsDbg
 
+#### Windows
+
 Once you have cloned the JsDbg repository, please follow the steps below to build the JsDbg source code.
 
 1. Open `jsdbg.sln` in Visual Studio 2017. The solution file is located in the `server` directory.
@@ -22,6 +24,21 @@ Once you have cloned the JsDbg repository, please follow the steps below to buil
 2. Build the solution by pressing `F6` or `Build -> Build Solution`.
       * Common build errors and fixes:
          * Missing Windows 10 SDK error - install the correct SDK version (should be specified in the error message) and rebuild the solution.
+
+#### Linux
+
+1. Get the [x64 .NET Core SDK binary](https://dotnet.microsoft.com/download) (click the x64 link in the left column in the table at the bottom). If you extract this into a different directory than `~/dotnet`, adjust the instructions below (and if you want to use `create_package.sh`, adjust the path to dotnet in this script).
+2. To build the server, run `~/dotnet/dotnet publish -c Release -r linux-x64` from the `server/JsDbg.Gdb` directory.
+3. The command will tell you where the output is; it should be in `bin/Release/netcoreapp2.1/linux-x64/publish` under the JsDbg.Gdb directory.
+4. Add the following code to your `~/.gdbinit`:
+```
+python
+import sys
+sys.path.insert(0, "<your home directory>/JsDbg/server/JsDbg.Gdb")
+import JsDbg
+end
+```
+5. If you cloned into `~/JsDbg` it should find the binary and extensions directory automatically; otherwise you will have to edit `JsDbg.py` and update the path in the `execPath` and `extensionsPath` variables in the `__init__` function of `JsDbg`.
 
 ### Running JsDbg
 
@@ -65,6 +82,22 @@ After you are able to successfully build JsDbg, please follow the steps below to
 7. Launch JsDbg by clicking the `JS` icon on the Debug toolbar or `Tools -> Launch JsDbg`. (Make sure you do this from the VS Experimental Instance and not the main copy of VS.)
 
 8. The JsDbg server will launch and prompt you to select a browser. If the debugger is attached to one browser, it is recommended that you choose a different browser for JsDbg.
+
+#### GDB
+
+1. Make sure you have added JsDbg to your `.gdbinit` as described in the Linux build instructions above.
+
+2. Type `jsdbg` at the gdb prompt to run it. It will automatically open a browser tab for JsDbg.
+
+3. Extensions should be automatically reloaded when you edit the source file; no need to restart the server. You can also reload the Python script without restarting if needed by entering the following code at the GDB prompt:
+```
+python
+import importlib
+importlib.reload(JsDbg)
+```
+
+4. For debugging JsDbg itself, you may want to set verbose to true, either by editing `JsDbg.py` or running `python JsDbg.jsdbg.verbose = True` at the gdb prompt.
+
 
 ## Components within JsDbg
 


docs/CONTRIBUTING.md
```diff
@@ -66,6 +80,20 @@ After you are able to successfully build JsDbg, please follow the steps below to
...
+After following the build instructions above, type `jsdbg` at the gdb prompt to run it. It will tell you the URL to open in your browser.
+
+Extensions should be automatically reloaded when you edit the source file; no need to restart the server. You can also reload the Python script without restarting if needed:
+```
+python
+import importlib
+importlib.reload(JsDbg)
+```
+
+For debugging, you may want to set verbose to true, either by editing `JsDbg.py` or running `python JsDbg.jsdbg.verbose = True` at the gdb prompt.
```
MSLaguana: By `debugging` here, are you referring to "using jsdbg to debug a program" or "debugging jsdbg itself"? I believe you are referring to the latter, but it'd be good to clear it up.

docs/CONTRIBUTING.md
```diff
@@ -66,6 +80,20 @@ After you are able to successfully build JsDbg, please follow the steps below to
 
 8. The JsDbg server will launch and prompt you to select a browser. If the debugger is attached to one browser, it is recommended that you choose a different browser for JsDbg.
 
+#### GDB
+
+After following the build instructions above, type `jsdbg` at the gdb prompt to run it. It will tell you the URL to open in your browser.
```
sanketj: I think we should clarify that the "build instructions" above are the Linux build instructions.

sanketj: Separately, from a UX point of view, it would be great if typing "jsdbg" would automatically open the URL in the browser (or pop the browser chooser). Is there a way to do this on Linux?

docs/CONTRIBUTING.md
```diff
@@ -23,6 +23,20 @@ Once you have cloned the JsDbg repository, please follow the steps below to buil
       * Common build errors and fixes:
          * Missing Windows 10 SDK error - install the correct SDK version (should be specified in the error message) and rebuild the solution.
 
+### Building JsDbg (Linux)
+
+1. Get the [x64 .NET Core SDK binary](https://dotnet.microsoft.com/download) (click the x64 link in the left column in the table at the bottom). Instructions below and `create_package.sh` assume you extract it into `~/dotnet`.
```
sanketj: Do you mean to say "Follow the instructions below"?

docs/CONTRIBUTING.md
```diff
@@ -23,6 +23,20 @@ Once you have cloned the JsDbg repository, please follow the steps below to buil
...
+2. To build the server, run `~/dotnet/dotnet publish -c Release -r linux-x64` from the `server/JsDbg.Gdb` directory.
+3. The command will tell you where the output is; it should be in `bin/Release/netcoreapp2.1/linux-x64/publish` under the JsDbg.Gdb directory
```
sanketj: nit: missing a '.' at the end

docs/CONTRIBUTING.md
sanketj: Also, not sure what you mean by "assume you extract it into ..."

docs/CONTRIBUTING.md
```diff
@@ -23,6 +23,20 @@ Once you have cloned the JsDbg repository, please follow the steps below to buil
...
+4. Add the `JsDbg.py` script to your `~/.gdbinit`. If you cloned into `~/JsDbg` it should find the binary and extensions directory automatically; otherwise you will have to edit the script and update the path. Code to add to `.gdbinit`:
+```
```
sanketj: This part isn't clear. Do we need to add both the code here and "JsDbg.py" to .gdbinit?

docs/CONTRIBUTING.md
```diff
@@ -66,6 +80,20 @@ After you are able to successfully build JsDbg, please follow the steps below to
 
 8. The JsDbg server will launch and prompt you to select a browser. If the debugger is attached to one browser, it is recommended that you choose a different browser for JsDbg.
 
+#### GDB
```
sanketj: Can we use the same numbering style instructions (1., 2., 3., ...) that we use for WinDbg and the VS debugger?

docs/CONTRIBUTING.md
```diff
@@ -13,7 +13,7 @@ Thank you for your interest in contributing to JsDbg! Code contributations, as w
 ### Getting the source
 The JsDbg source can be obtained from the [JsDbg repository on GitHub](https://aka.ms/jsdbg). The URL for cloning the repository is: `https://github.com/MicrosoftEdge/JsDbg.git`.
 
-### Building JsDbg
+### Building JsDbg (Windows)
```
sanketj: Instead of "Building JsDbg (Windows)" and "Building JsDbg (Linux)", can we keep "Building JsDbg" as is and add subheadings for 'Windows" and "Linux"?

cbiesinger: Done

docs/CONTRIBUTING.md
cbiesinger: I meant that the instructions were written with an assumption that dotnet is extracted into ~/dotnet. I've rewritten this a bit.

docs/CONTRIBUTING.md
cbiesinger: done

docs/CONTRIBUTING.md
cbiesinger: Good point, I've rewritten this.

docs/CONTRIBUTING.md
cbiesinger: done

docs/CONTRIBUTING.md
cbiesinger: Done

cbiesinger: https://github.com/MicrosoftEdge/JsDbg/pull/97 does that

docs/CONTRIBUTING.md
cbiesinger: Clarified that

docs/CONTRIBUTING.md
```diff
@@ -23,6 +25,21 @@ Once you have cloned the JsDbg repository, please follow the steps below to buil
       * Common build errors and fixes:
          * Missing Windows 10 SDK error - install the correct SDK version (should be specified in the error message) and rebuild the solution.
 
+#### Linux
+
+1. Get the [x64 .NET Core SDK binary](https://dotnet.microsoft.com/download) (click the x64 link in the left column in the table at the bottom). If you extract this into a different directory than `~/dotnet`, adjust the instructions below (and if you want to use `create_package.sh`, adjust the path to dotnet in this script)
```
sanketj: nit: missing a '.' at the end

docs/CONTRIBUTING.md
```diff
@@ -23,6 +25,21 @@ Once you have cloned the JsDbg repository, please follow the steps below to buil
...
+3. The command will tell you where the output is; it should be in `bin/Release/netcoreapp2.1/linux-x64/publish` under the JsDbg.Gdb directory.
+4. Add the following code to your `~/.gdbinit`:
+```
+python
+import sys
+sys.path.insert(0, "<your home directory>/JsDbg/server/JsDbg.Gdb")
+import JsDbg
+end
+```
+5. If you cloned into `~/JsDbg` it should find the binary and extensions directory automatically; otherwise you will have to edit `JsDbg.py` and update the path.
```
sanketj: I think "update the path" is a little unclear here. Could you make this part a little more explicit?

docs/CONTRIBUTING.md
```diff
@@ -66,6 +83,22 @@ After you are able to successfully build JsDbg, please follow the steps below to
 
 8. The JsDbg server will launch and prompt you to select a browser. If the debugger is attached to one browser, it is recommended that you choose a different browser for JsDbg.
 
+#### GDB
+
+1. Make sure you added JsDbg to your `.gdbinit` as described in the Linux build instructions above
```
sanketj: you "have" added

also missing a '.' at the end

docs/CONTRIBUTING.md
```diff
@@ -66,6 +83,22 @@ After you are able to successfully build JsDbg, please follow the steps below to
...
+
+2. Type `jsdbg` at the gdb prompt to run it. It will automatically open a browser tab for JsDbg.
+
+3. Extensions should be automatically reloaded when you edit the source file; no need to restart the server. You can also reload the Python script without restarting if needed:
```
sanketj: At the end of the second sentence (about reloading without restart), I think we need to add something along the lines of "by running the code below". Also, would be good to clarify where to run the code (gdb prompt? shell?).

docs/CONTRIBUTING.md
cbiesinger: Done

docs/CONTRIBUTING.md
cbiesinger: Done

docs/CONTRIBUTING.md
cbiesinger: Done

docs/CONTRIBUTING.md
cbiesinger: Done

PR #97, opened 2019-05-08
[gdb] Automatically open a browser when starting jsdbg

cbiesinger:


server/JsDbg.Gdb/JsDbg.py
```diff
@@ -185,6 +186,11 @@ def __repr__(self):
         return '{%s#%d}' % (self.name, self.baseAddress)
 
 
+def ServerStarted(url):
+    print('Opening browser for %s' % (url))
+    webbrowser.open_new_tab(url)
```
sanketj: Does this work if we're attached to the browser process of the default web browser?

cbiesinger: No, but that's uncommon on Linux. I don't think I can install debug symbols for it myself (though some linux distributions can). Also I'm not sure how to get a browser chooser on Linux.

Worst case, the URL does get printed so the user can manually load it in a different browser.

sanketj: I think the same problem occurs with the GPU process as well. If it's not possible to show the browser chooser, perhaps we can add something next to the URL that gets printed? Perhaps something like: if you're debugging your default browser, you may need to use a different browser to open JsDbg.

cbiesinger: I'm not sure I understand the relevance of the GPU process, but I did make https://github.com/MicrosoftEdge/JsDbg/pull/107 to expand on the message.

sanketj: I mentioned the GPU process just to make the point that it may not be that uncommon to encounter this issue. Anyways, thanks for adding the message! :)

PR #98, opened 2019-05-08
Support std::__Cr::unordered_map and std::__Cr::vector

cbiesinger:
And make ax-tree use the extended field instead of directly accessing
the hashtable, so that it also works with a different STL implementation.

Note, this does not quite make ax-tree work on Linux yet, still figuring
out why.

### DIFF
--- a/extensions/target-specific/stl-helpers/stl-helpers.js
+++ b/extensions/target-specific/stl-helpers/stl-helpers.js
@@ -33,6 +33,28 @@ DbgObject.AddArrayField(
     }
 );
 
+DbgObject.AddArrayField(
+    (type) => {
+        return type.name().match(/^std::__Cr::vector<(.*)>$/) != null;
+    },
+    "Elements",
+    (type) => {
+        return type.templateParameters()[0];
+    },
+    (vector) => {
+        return vector.f("__begin_").array(vector.desc("Size"));
+    }
+);
+
+function computeVectorSize(firstElement, lastElement, elementSize) {
+    if (!firstElement.isNull()) {
+        console.assert(!lastElement.isNull());
+        return lastElement.pointerValue().minus(firstElement.pointerValue()).divide(elementSize);
+    } else {
+        return 0;
+    }
+}
+
 DbgObject.AddTypeDescription(
     (type) => {
         return type.name().match(/^std::vector<(.*)>$/) != null;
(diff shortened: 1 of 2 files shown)


extensions/target-specific/stl-helpers/stl-helpers.js
```diff
@@ -53,6 +67,25 @@ DbgObject.AddTypeDescription(
...
+DbgObject.AddTypeDescription(
+    (type) => {
+        return type.name().match(/^std::__Cr::vector<(.*)>$/) != null;
+    },
+    "Size",
+    false,
+    (vector) => {
+        return Promise.all([vector.f("__begin_"), vector.f("__end_"), vector.f("__begin_").size()])
+        .thenAll((firstElement, lastElement, elementSize) => {
+            if (!firstElement.isNull()) {
```
sanketj: Consider sharing the code inside .thenAll between this and std::vector's Size type description?

extensions/target-specific/stl-helpers/stl-helpers.js
```diff
@@ -151,6 +184,25 @@ function inOrderTraversal(rootNodeOrPromise, leftField, rightField, valueField,
...
+    },
+    (map) => {
+        return map.f("__table_").then((table) => {
+            var fromType = map.type.templateParameters()[0];
+            var toType = map.type.templateParameters()[1];
+            var nodeTypeName = `std::__Cr::__hash_node<std::__Cr::__hash_value_type<${fromType}, ${toType}>, void*>*`;
+            var nodeType = DbgObjectType(nodeTypeName, table.type);
+            return map.f("__table_.__p1_.__value_").list("__next_").map(
+                (elem) => elem.as(nodeType).deref().f("__value_.__cc"));
+    })}
```
sanketj: nit: indentation is off here

extensions/target-specific/stl-helpers/stl-helpers.js
cbiesinger: Will do before merging

cbiesinger: Done

extensions/target-specific/stl-helpers/stl-helpers.js
cbiesinger: Done

PR #100, opened 2019-05-09
[gdb] Let gdb handle converting types to strings

cbiesinger:
It is difficult to get them right. In particular, this makes us
handle char[] correctly.

### DIFF
--- a/server/JsDbg.Gdb/JsDbg.py
+++ b/server/JsDbg.Gdb/JsDbg.py
@@ -101,18 +101,19 @@ def DebuggerQuery(tag, command):
         return "%d!%s" % (tag, str(err[1]))
 
 
-def FormatType(symbol_type):
-        pointer_depth = 0
-        t = symbol_type.strip_typedefs()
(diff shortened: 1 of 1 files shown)


server/JsDbg.Gdb/JsDbg.py
```diff
@@ -102,17 +102,12 @@ def DebuggerQuery(tag, command):
 
 
 def FormatType(symbol_type):
-        pointer_depth = 0
         t = symbol_type.strip_typedefs()
-        while t.code == gdb.TYPE_CODE_PTR or t.code == gdb.TYPE_CODE_ARRAY:
-            pointer_depth = pointer_depth + 1
-            t = t.target()
-
         if (t.code == gdb.TYPE_CODE_FUNC):
```
MSLaguana: Does this work with function pointers, or arrays of/pointers to function pointers?

I know that when I was first writing this, function pointers were a pain to round-trip, and if you don't strip off the pointers/arrays you won't know whether it is a function pointer

cbiesinger: OK, that should be fixed now

PR #104, opened 2019-05-09
[gdb] Support anonymous structs as well as unions

cbiesinger:
This is required to support std::basic_string

### DIFF
--- a/server/JsDbg.Gdb/JsDbg.py
+++ b/server/JsDbg.Gdb/JsDbg.py
@@ -215,11 +215,12 @@ def GetAllFields(module, type, includeBaseTypes):
             continue
 
         if not field.name:
-            if field.type.code != gdb.TYPE_CODE_UNION:
+            if field.type.code != gdb.TYPE_CODE_UNION and field.type.code != gdb.TYPE_CODE_STRUCT:
                 # Don't know how to handle this
                 continue
 
-            resultFields.extend([SFieldResult(f, field.bitpos) for f in field.type.fields()])
+            resultFields.extend([SFieldResult(f, field.bitpos)
+                for f in field.type.fields() if not f.is_base_class])
             continue
 
         resultFields.append(SFieldResult(field))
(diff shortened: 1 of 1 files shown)


server/JsDbg.Gdb/JsDbg.py
```diff
@@ -251,9 +251,10 @@ def LookupField(module, type, field):
...
-        # Handle anonymous unions. They are a bit tricky because we have
-        # to recurse into their fields but keep track of their offset.
-        unions = [u for u in fields if not u.name and u.type.code == gdb.TYPE_CODE_UNION]
+        # Handle anonymous unions and structs. They are a bit tricky because we
+        #have to recurse into their fields but keep track of their offset.
+        unions = [u for u in fields if not u.name and
+          (u.type.code == gdb.TYPE_CODE_UNION or u.type.code == gdb.TYPE_CODE_STRUCT)]
         for union in unions:
             for f in union.type.fields():
                 if f.name == field:
```
MSLaguana: I think you'll need to update the `union.bitpos` on the following line; all members of a union start at the same offset, but members of a struct have to increment their offset. I think this will mean you may have to handle them separately?

MSLaguana: Hmm, maybe I'm wrong actually and it does get handled... the `union.bitpos` is the additional offset to apply, so it'll be the offset of the containing struct in the anonymous struct case; that sounds plausible... Could you update the comment on `SFieldResult` to reference "anonymous containers" or something rather than just "anomymous unions"?

cbiesinger: Exactly, that's why this works,

Updated the comment and made another couple of minor changes, please take another look.

server/JsDbg.Gdb/JsDbg.py
```diff
@@ -251,13 +252,14 @@ def LookupField(module, type, field):
         if match:
             return SFieldResult(match[0])
 
-        # Handle anonymous unions. They are a bit tricky because we have
-        # to recurse into their fields but keep track of their offset.
-        unions = [u for u in fields if not u.name and u.type.code == gdb.TYPE_CODE_UNION]
-        for union in unions:
-            for f in union.type.fields():
+        # Handle anonymous unions and structs. They are a bit tricky because we
+        #have to recurse into their fields but keep track of their offset.
```
MSLaguana: Small typo here

cbiesinger: Thanks, fixed

PR #105, opened 2019-05-09
[stl] Support libcxx pair and basic_string

cbiesinger:
String printer based on
https://cs.chromium.org/chromium/src/third_party/libcxx-pretty-printers/printers.py

### DIFF
--- a/extensions/target-specific/stl-helpers/stl-helpers.js
+++ b/extensions/target-specific/stl-helpers/stl-helpers.js
@@ -260,7 +260,27 @@ DbgObject.AddTypeDescription(
 );
 
 DbgObject.AddTypeDescription(
-    (type) => type.name().match(/^std::pair<.*>$/) != null,
+    (type) => type.name().match(/^std::__(Cr|1)::basic_string<.*>$/) != null,
+    "Text",
+    true,
+    (str) => {
+        var ss = str.f("__r_.__value_.__s");
+        var shortMask = DbgObject.constantValue(str.type, "__short_mask");
+        var size = ss.f("__size_").uval();
+        return Promise.all([ss, shortMask, size]).thenAll((ss, shortMask, size) => {
+            if ((size & shortMask) == 0) {
+                var len = (shortMask == 1) ? size >> 1 : size;
+                return ss.f("__data_").string(len);
+            } else {
+                var sl = str.f("__r_.__value_.__l");
+                return sl.f("__data_").string(sl.f("__size_"));
+            }
+        });
+    }
+);
+
+DbgObject.AddTypeDescription(
+    (type) => type.name().match(/^std::(__(Cr|1)::)?pair<.*>$/) != null,
     "Pair",
     true,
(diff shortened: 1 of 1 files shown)


extensions/target-specific/stl-helpers/stl-helpers.js
```diff
@@ -260,7 +260,27 @@ DbgObject.AddTypeDescription(
 );
 
 DbgObject.AddTypeDescription(
-    (type) => type.name().match(/^std::pair<.*>$/) != null,
+    (type) => type.name().match(/^std::__Cr::basic_string<.*>$/) != null,
```
sanketj: Is "std::__Cr" the only STL prefix that needs to be supported? What about "std::__1"? @KurtCattiSchmidt noticed that the AXTree extension was broken because some of the STL types being used have changed to "std::__1".  From looking online, it seems that "__1" is an inline namespace: https://stackoverflow.com/questions/29293394/where-does-the-1-symbol-come-from-when-using-llvms-libc, but I'm not sure why we're seeing this now. (Wonder if there has been a STL or compiler update.) Do you know anything about this? Can you and Kurt sync about this to make sure we handle all these cases correctly.

cbiesinger: Huh, I don't know anything about std::__1. @KurtCattiSchmidt , was that on windows/linux/mac? any special compile flags?

cbiesinger: __Cr comes from here: https://cs.chromium.org/chromium/src/build/config/c%2B%2B/BUILD.gn?type=cs&q=_LIBCPP_ABI_VERSION&sq=package:chromium&g=0&l=50

It looks like __1 is from here: https://cs.chromium.org/chromium/src/buildtools/third_party/libc%2B%2B/trunk/include/__config?type=cs&q=_LIBCPP_ABI_VERSION&sq=package:chromium&g=0&l=38

So it seems like we need to support __1 as well, for non-component builds. I think it'd be better to do this in a separate change because that also affects the other std::__Cr occurences in this file.

cbiesinger: OK, I made https://github.com/MicrosoftEdge/JsDbg/pull/109 and also updated this pull request to allow both __Cr and __1.

KurtCattiSchmidt: Thank you, I'll take a look at your PR and see if it fixes my issue. I'm running Win10, with a debug component build.

cbiesinger: Looks like Windows was switched to libc++ in https://chromium.googlesource.com/chromium/src/+/4b298260f5327c09983083055a5775d5ce719cc1, so very recently.

You'll want to apply this pull request as well, in addition to #109

sanketj: Thanks @cbiesinger for finding the PR that changed this, and for fixing the stl helpers. I'll take a look at that PR shortly, but I'm good with fixing that separately.

PR #110, opened 2019-05-10
[gdb] Format pointers in hex for Python

cbiesinger:
No behavior change but makes debugging JsDbg easier.

### DIFF
--- a/server/JsDbg.Gdb/GdbDebugger.cs
+++ b/server/JsDbg.Gdb/GdbDebugger.cs
@@ -410,7 +410,7 @@ namespace JsDbg.Gdb {
             // '0x4004f1 <twiddle<int []>(int)+17>'
             // which has the same informationas -var-create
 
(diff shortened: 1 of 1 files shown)


server/JsDbg.Gdb/GdbDebugger.cs
```diff
@@ -423,7 +423,7 @@ class GdbDebugger : IDebugger {
             int lastPlus = response.LastIndexOf('+');
             int lastGreater = response.LastIndexOf('>');
             if (firstLess < 0 || lastGreater < 0) {
-                throw new DebuggerException(String.Format("Address {0} is not a symbol", pointer));
+                throw new DebuggerException(String.Format("Address {0:x} is not a symbol", pointer));
```
sanketj: Do we not want to show the "0x" here?

cbiesinger: Fixed, thanks

PR #119, opened 2019-05-14
[memorycache] Support unaligned memory access

cbiesinger:
Let me know if this indentation of the second line of the if is OK

### DIFF
--- a/extensions/memorycache/memorycache.js
+++ b/extensions/memorycache/memorycache.js
@@ -50,6 +50,10 @@ Loader.OnLoad(function() {
         return address.mod(PAGE_SIZE).divide(size);
     }
 
+    function isUnaligned(address, size) {
+        return !address.mod(size).isZero();
+    }
+
     function Uint64Viewer(arrayBuffer) {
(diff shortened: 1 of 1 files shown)


extensions/memorycache/memorycache.js
```diff
@@ -149,7 +153,8 @@ Loader.OnLoad(function() {
         address = new PointerMath.Pointer(address).value();
 
         var viewer = getArrayViewer(size, isUnsigned, isFloat);
-        if (!loadPage(getPage(address), viewer, function(view) {
+        if (isUnaligned(address, size) ||
```
sanketj: Can we add a comment or two about this check and why it is needed? Both here and below.

cbiesinger: Done

sanketj: LGTM. Please get sign off from @petersalas as well.

cbiesinger:
I'm sorry, I saw your comment too late :(

@petersalas , does this look OK to you? If not I can revert/fix

petersalas:
LGTM!

PR #121, opened 2019-05-15
Make "chrome" a multi-process module

cbiesinger:
On Linux, Chrome does not use chrome_child; move chrome to multi-process
modules so that blink-helpers and wtf-helpers work correctly.

### DIFF
--- a/extensions/target-specific/chromium/chromium-helpers/chromium-helpers.js
+++ b/extensions/target-specific/chromium/chromium-helpers/chromium-helpers.js
@@ -82,7 +82,7 @@ var Chromium = null;
             _help_BrowserProcessEquivalentModules: {
                 description: "List of modules used in the Chromium browser process.",
             },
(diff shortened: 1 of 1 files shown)


extensions/target-specific/chromium/chromium-helpers/chromium-helpers.js
```diff
@@ -82,7 +82,7 @@ var Chromium = null;
             _help_BrowserProcessEquivalentModules: {
                 description: "List of modules used in the Chromium browser process.",
             },
-            BrowserProcessEquivalentModules: ["msedge", "accessibility", "chrome"],
+            BrowserProcessEquivalentModules: ["msedge", "accessibility"],
```
sanketj: "msedge" is basically a rename of "chrome", so we should move that over as well.

cbiesinger: Done

PR #123, opened 2019-05-15
JsDbg.VisualStudio: Add JsDbg support for VS 2019

sanketj:
To support building JsDbg.VisualStudio on both VS 2017 and 2019, changed  "Microsoft.VisualStudio.Shell.Framework" from version 15 to a reference to the latest version (16.0.28729) of the "Microsoft.VisualStudio.Shell.Framework" NuGet package. Installing that NuGet package introduces the rest of the references being added here.

To enable the JsDbg vsix to work on VS 2019, changed the installation and prerequisite targets to include the latest version of VS.

### DIFF
--- a/server/JsDbg.VisualStudio/JsDbg.VisualStudio.csproj
+++ b/server/JsDbg.VisualStudio/JsDbg.VisualStudio.csproj
@@ -91,6 +91,9 @@
     <Reference Include="Microsoft.VisualStudio.CommandBars, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
       <EmbedInteropTypes>True</EmbedInteropTypes>
     </Reference>
+    <Reference Include="Microsoft.VisualStudio.CoreUtility, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\packages\Microsoft.VisualStudio.CoreUtility.15.6.27740\lib\net46\Microsoft.VisualStudio.CoreUtility.dll</HintPath>
+    </Reference>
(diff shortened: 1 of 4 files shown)


sanketj:
@petersalas @cbiesinger @sushraja-msft Could one of you please check if you're able to build JsDbg.VisualStudio successfully with these changes?

cbiesinger:
Yes, I can still build on VS 2017.

server/JsDbg.VisualStudio/JsDbg.VisualStudio.csproj
```diff
@@ -115,13 +122,70 @@
...
+    </Reference>
+    <Reference Include="System.ComponentModel.Composition" />
     <Reference Include="System.Core" />
     <Reference Include="System.Data" />
     <Reference Include="System.Design" />
     <Reference Include="System.Drawing" />
+    <Reference Include="System.IO, Version=4.1.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\packages\System.IO.4.3.0\lib\net462\System.IO.dll</HintPath>
+      <Private>True</Private>
+      <Private>True</Private>
```
cbiesinger: Weird that there are two <Private> tags here?

sanketj: Yeah, that is weird. Didn't notice that. But looks like all the private references have two tags. I did some research online about this but couldn't find anything. Will keep digging separately and if I figure out why, I'll submit a fix.

sanketj:
> I don't understand all the changes to the project file but it still compiles for me, so lgtm

These are dependencies that have come along for the ride after taking a dependency on the "Microsoft.VisualStudio.Shell.Framework" NuGet package.

PR #128, opened 2019-05-16
[gdb] Make GetBaseTypes faster

cbiesinger:
No need to call gdb.lookup_type for every base class.

### DIFF
--- a/server/JsDbg.Gdb/JsDbg.py
+++ b/server/JsDbg.Gdb/JsDbg.py
@@ -245,24 +245,34 @@ def GetAllFields(module, type, includeBaseTypes):
 
     return resultFields
 
+
 # extra_bitoffset is used when we call this function recursively in multiple
 # inheritance cases.
-def GetBaseTypes(module, type, extra_bitoffset=0):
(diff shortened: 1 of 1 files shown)


server/JsDbg.Gdb/JsDbg.py
```diff
@@ -245,25 +245,35 @@ def GetAllFields(module, type, includeBaseTypes):
...
 
     return resultFields
 
+
+def GetBaseTypes(module, type):
+    try:
+        t = gdb.lookup_type(type)
+    except:
+        # Type is a base type?
+        return [SBaseTypeResult(module, type, 0)]
```
sanketj: Shouldn't the second parameter be type.name?

cbiesinger: In this specific location, type is the name of the type as passed in from GdbDebugger.cs. But since that's a bit subtle, I renamed the parameter to make it more clear.

sanketj: Oops sorry...I completely missed this change last week.

PR #130, opened 2019-05-20
Show an example WinDbg commandline to start a server.

cbiesinger:
Makes it a bit easier for users.


server/JsDbg.WinDbg/Program.cs
```diff
@@ -32,6 +32,7 @@ public class Program {
             string remoteString;
             if (args.Length < 1 || args[0] == "/ask") {
                 // A debugger string wasn't specified.  Prompt for a debug string instead.
+                Console.Write("Hint: You can use '.server npipe:pipe=jsdbg' in WinDbg.")
```
sanketj: might be better to say "You can run the following command to create a remote debugging server in WinDbg: '.server npipe:pipe=jsdbg'"

cbiesinger: Done

PR #134, opened 2019-05-22
Handle references and volatile in DbgObjectType

cbiesinger:
These can be returned by GDB. volatile can be ignored, and
references should be converted to pointers.

### DIFF
--- a/extensions/dbgobject/type/dbgobject-type.js
+++ b/extensions/dbgobject/type/dbgobject-type.js
@@ -26,7 +26,8 @@ Loader.OnLoad(function () {
 
         // Normalize type name.
         this._name = name.trim();  // strip outer whitespace
-        this._name = this._name.replace(/^const\s+/, "");  // remove leading "const" and whitespace
+        this._name = this._name.replace(/&/g, "*"); // treat references like pointers
+        this._name = this._name.replace(/^((const|volatile)\s+)+/g, "");  // remove leading "const" and whitespace
         this._name = this._name.replace(/\s+const$/, "");  // remove trailing "const" and whitespace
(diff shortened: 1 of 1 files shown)


extensions/dbgobject/type/dbgobject-type.js
```diff
@@ -26,7 +26,8 @@ Loader.OnLoad(function () {
 
         // Normalize type name.
         this._name = name.trim();  // strip outer whitespace
-        this._name = this._name.replace(/^const\s+/, "");  // remove leading "const" and whitespace
+        this._name = this._name.replace(/&/g, "*"); // treat references like pointers
+        this._name = this._name.replace(/^((const|volatile)\s+)+/g, "");  // remove leading "const" and whitespace
```
sanketj: I think technically "const volatile" is possible too. :)

sanketj: Should we also handle "mutable" here?

cbiesinger: Yes, the last + in the regex and the /g handle that case.

cbiesinger: Yes, the last + in the regex and the /g handle that case.

cbiesinger: Hmm, GDB does not consider "mutable" to be part of the type. And I guess it isn't really part of the type -- more part of the field declaration? So I think it's fine to leave as-is.

sanketj: Missed that, sorry. That looks good.

sanketj: Ok, good to know. This is fine in that case.

PR #137, opened 2019-05-23
[gdb] Make GetSymbolsInStackFrame work in frames without symbols

cbiesinger:
If we don't have symbols for a frame, block() throws an exception. Instead,
just return an empty list.

### DIFF
--- a/server/JsDbg.Gdb/JsDbg.py
+++ b/server/JsDbg.Gdb/JsDbg.py
@@ -344,7 +344,11 @@ def GetSymbolsInStackFrame(instructionAddress, stackAddress, frameAddress):
         frame = frame.older()
 
     if frame:
-        block = frame.block()
+        try:
+            block = frame.block()
+        except:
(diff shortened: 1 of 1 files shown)


server/JsDbg.Gdb/JsDbg.py
```diff
@@ -344,7 +344,11 @@ def GetSymbolsInStackFrame(instructionAddress, stackAddress, frameAddress):
         frame = frame.older()
 
     if frame:
-        block = frame.block()
+        try:
+            block = frame.block()
+        except:
+            # We are probably missing symbols for this frame.
```
sanketj: Will we end up showing "??" for this case?

cbiesinger: No, for this case we'd just show no local variables.

sanketj: Ok, thanks for the clarification.

PR #139, opened 2019-06-03
[gdb] Also accept a different name for rr binaries

cbiesinger:
When using "rr pack" rr uses a different name for binaries/libraries,
accept that as well.

### DIFF
--- a/server/JsDbg.Gdb/JsDbg.py
+++ b/server/JsDbg.Gdb/JsDbg.py
@@ -309,10 +309,10 @@ def LookupGlobalSymbol(module, symbol):
 
 
 def GetModuleForName(module):
-  # If we are running under rr, it renames (hardlinks) the executable to
-  # mmap_hardlink_N_foo; allow for that.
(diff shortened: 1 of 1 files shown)


server/JsDbg.Gdb/JsDbg.py
```diff
@@ -309,10 +309,10 @@ def LookupGlobalSymbol(module, symbol):
 
 
 def GetModuleForName(module):
-  # If we are running under rr, it renames (hardlinks) the executable to
-  # mmap_hardlink_N_foo; allow for that.
-  matches = re.match("^.*/mmap_hardlink_[0-9]+_(.*)$", gdb.current_progspace().filename)
-  if matches and matches.groups()[0] == module:
+  # If we are running under rr, it renames/hardlinks/copy the executable to
```
dandclark: renames/hardlinks/copy ==> renames/hardlinks/copies?

cbiesinger: Indeed; fixed now

PR #141, opened 2019-06-04
[gdb] Have GetSymbolsInStackFrame return the right module

cbiesinger:
Using the module for the pc address matches what DiaDebugger does.
(Should it use IDiaSymbol::get_libraryName instead?)

### DIFF
--- a/server/JsDbg.Gdb/GdbDebugger.cs
+++ b/server/JsDbg.Gdb/GdbDebugger.cs
@@ -317,13 +317,13 @@ namespace JsDbg.Gdb {
             foreach (string symString in objects) {
                 // '{%s#%d#%s}' % (self.name, self.symbolResult.pointer, s.symbolResult.type)
                 string[] properties = symString.Split("#");
-                Debug.Assert(properties.Length == 3);
+                Debug.Assert(properties.Length == 4);
                 SNamedSymbol sym = new SNamedSymbol();
-                sym.Name = properties[0];
+                sym.Name = properties[1];
                 sym.Symbol = new SSymbolResult();
-                sym.Symbol.Pointer = UInt64.Parse(properties[1]);
-                sym.Symbol.Module = "N/A";
-                sym.Symbol.Type = properties[2];
+                sym.Symbol.Pointer = UInt64.Parse(properties[2]);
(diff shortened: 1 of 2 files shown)


MSLaguana:
This change seems reasonable to me. Do you know whether it allows proper disambiguation of things which come from different modules? Or is that not supported?

cbiesinger:
I'm not sure I understand your question correctly but here are some thoughts:
- DiaDebugger returns the module for the PC address just like this code does, so if the type of the variable is declared in the same module as the function this is all good. (I'm not sure that code is really ideal though...):
https://github.com/MicrosoftEdge/JsDbg/blob/master/server/JsDbg.Windows/DiaDebugger.cs#L351
https://github.com/MicrosoftEdge/JsDbg/blob/master/server/JsDbg.Windows/DiaDebugger.cs#L399

- Maybe you're asking if we should return the module that the type was declared in? If so, we can do that in very new GDBs that have my patch from https://sourceware.org/ml/gdb-patches/2019-06/msg00063.html but this pull request does not make use of that

MSLaguana:
Perhaps my question was coming from too much of a place of ignorance :)

I had thought that the module was so that `foo!sym` could be disambiguated from `bar!sym`, and I wasn't sure if that worked out with these changes (e.g. if there are required changes elsewhere when the module is passed in that need to be accounted for in the python code)

MSLaguana:
E.g. if `GetAllFields("foo", "sym", false)` and `GetAllFields("bar", "sym", false)` are capable of returning different results since they refer to different types.

cbiesinger:
Oh -- well that is a different function, but as far as I know gdb does not support disambiguating that, at least in the python API

MSLaguana:
Yeah it is a different function, but I wasn't sure if that may now come up since the debugger will return modules other than `"N/A"`

cbiesinger:
Well, various functions already returned non-N/A modules. LookupSymbolName returns a fully correct module, and other functions return the module that was passed in.

But yeah, don't think this is supported. Maybe I could extend `gdb.lookup_type` to take an `objfile` argument...

cbiesinger:
(Or more likely, give `gdb.Objfile` a `lookup_type` function)

PR #143, opened 2019-06-10
Adding more (Cr|1) matches and handling space differences in unordered_map

KurtCattiSchmidt:
AXTree isn't currently working on Windows due to additional types expecting __Cr as the STL prefix instead of __1. This change updates them all to be equivalent.

This change also fixes another issue with spacing in unordered_map. The solution is not ideal, but it unblocks AXTree functionality on Windows. I will sync with Sanket when he gets back about a cleaner long-term solution. I had considered removing spaces all from `this._name` in `extensions\dbgobject\type\dbgobject-type.js`, but that caused issues when it attempts to look-up later. One option is a normalized_name_ and an original name_, using the normalized_name_ for matching and the canonical name_ for binary lookup.


cbiesinger:
Except for the space change -- if windows needs the space to be removed, you can just remove it, gdb doesn't care either way.

KurtCattiSchmidt:
> I'm confused, master already has these changes as far as I can tell?

You're right, I was on a very old master

> Except for the space change -- if windows needs the space to be removed, you can just remove it, gdb doesn't care either way.

I wasn't aware of that, and it simplifies my change quite a bit.

When I pulled/rebased, it messed up the diff so I'll just abandon this PR and create a new one with just the spacing issues addressed.

PR #144, opened 2019-06-10
Adjusting spaces to match Windows binary expectations

KurtCattiSchmidt:
This fixes the AXTree viewer on Windows.

Based on @cbiesinger 's comment in #143, the spacing between parameters and pointer *'s doesn't matter with gdb. However, with WinDBG they do matter, so this change adjusts the spacing to match WinDBG's expectations.


cbiesinger:
I don't really understand why this whitespace is significant to windbg, but clearly it is:
```
0:042> dt content!std::__1::__value_type<blink::WebFrame *,content::RenderFrameImpl *>
   +0x000 __cc             : std::__1::pair<blink::WebFrame *const,content::RenderFrameImpl *>
0:042> dt content!std::__1::__value_type<blink::WebFrame *, content::RenderFrameImpl *>
Symbol content!std::__1::__value_type<blink::WebFrame *, content::RenderFrameImpl *> not found.
```

PR #148, opened 2019-06-14
[gdb] Add makefiles for easier compiling

cbiesinger:


server/JsDbg.Gdb/Makefile
```diff
@@ -0,0 +1,28 @@
+PREFIX=/usr/local
+DOTNET=$(HOME)/dotnet/dotnet
+
+PUBLISH=bin/Release/netcoreapp2.1/linux-x64/publish
```
dlibby-: Is `netcoreapp2.1` a constant, or should we get it from the version somehow? Looks like the current download page points to 2.2 - does it still produce output at netcoreapp2.1?

Makefile
```diff
@@ -0,0 +1,5 @@
+all:
```
dlibby-: My Makefile knowledge is very limited (so this is more for my understanding) - do we need this all target or would the % one take care of this as well?

server/JsDbg.Gdb/Makefile
dlibby-: Looks like this matches the existing path from the python script, so I'm guessing not an issue (or at least can be tackled in a separate change).

cbiesinger: Fairly sure this comes from the TargetFramework in the .csproj file -- I do have the dotnet sdk 2.2 and it does output into netcoreapp2.1.

Makefile
cbiesinger: Unfortunately it seems we need this target if we want "make" without an explicit target to work.

Issue #126, opened 2019-05-16
Allow truncating path to variables


cbiesinger:
Today I wanted to show a member variable in the tree with a fairly long path:
Layer.scrollable_area_.raw_.scrollbar_manager_.v_bar_is_attached_:false

This takes up a lot of horizontal space... can we allow truncating them? Or always truncate in the middle when they get long enough, or something?

sanketj:
Interesting idea!

For long paths, we could do something like: add a "..." in the middle and the show the full path when the user hits alt+click.

The question though is when do we decide to show the "...". After we expand through a certain number of fields? After we cross an arbitrary path size? Or would it have to vary based on window size? I'm not sure what the right answer is here.

sanketj:
@petersalas , have you ever thought about this?

petersalas:
Yeah, it can definitely be problem, especially when multiple fields are being showed and it's embarrassingly bad when it's `a.really.long.sequence.of.objects.ending.in.fieldA` and `a.really.long.sequence.of.objects.ending.in.fieldB`.

The ideas around trimming sound good to me...maybe elide fields between 2 and N-1?

sanketj:
@cbiesinger @petersalas , do we still want this now that we have #155 ?

cbiesinger:
You could certainly make an argument either way but I'm leaning towards keeping this open because it would be helpful to conserve screen width when showing multiple variables with long paths

PR #155, opened 2019-06-21
Show the variable path more subtly

cbiesinger:
When showing multiple variables in the tree display, this
makes it easier to scan the display for the value vs the
path.

Bug: #126

### DIFF
--- a/extensions/tree-inspector/field-selector/field-selector.css
+++ b/extensions/tree-inspector/field-selector/field-selector.css
@@ -68,6 +68,14 @@
     display:block;
 }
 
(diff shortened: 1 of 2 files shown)


cbiesinger:
Looks like this. Thoughts? Not exactly what bug #126 suggested but in a similar vein.
Screenshot from 2019-06-21 14-56-47

petersalas:
I like it!  In a sense, the last field _is_ the thing, and the other ones are _the relationship_ between the object and "the thing" so differentiating between them makes sense to me.

PR #156, opened 2019-06-21
[gdb] Use %s to format the result from GetTargetProcess

cbiesinger:
It returns a string now. Fixes an error in Python 2.

### DIFF
--- a/server/JsDbg.Gdb/JsDbg.py
+++ b/server/JsDbg.Gdb/JsDbg.py
@@ -503,7 +503,7 @@ def CheckForProcessAndThreadChange():
         current_thread = None
 
     if last_pid != current_process and jsdbg:
-        jsdbg.SendGdbEvent('proc %i' % (current_process))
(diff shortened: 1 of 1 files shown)


server/JsDbg.Gdb/JsDbg.py
```diff
@@ -503,7 +503,7 @@ def CheckForProcessAndThreadChange():
         current_thread = None
 
     if last_pid != current_process and jsdbg:
-        jsdbg.SendGdbEvent('proc %i' % (current_process))
+        jsdbg.SendGdbEvent('proc %s' % (current_process))
     if last_tid != current_thread and jsdbg:
         jsdbg.SendGdbEvent('thread %i' % (current_thread))
```
sanketj: Does current_thread not have the same problem?

cbiesinger: No, because for GetTargetThread we return a number. See line 477. (We return a string for GetTargetProcess because some versions of gdb/python otherwise serialize the number badly)

PR #160, opened 2019-06-25
[gdb] Add a unit test for JsDbg.FormatModule

cbiesinger:


server/JsDbg.Gdb/JsDbg_test.py
```diff
@@ -0,0 +1,24 @@
...
+    COMMAND_USER = 1
+
+    class Command(object):
+        def __init__(self, name, type):
+            pass
+
+sys.modules['gdb'] = GdbModule
+import JsDbg
+
+class TestJsDbg(unittest.TestCase):
```
sanketj: Would be good to add some instructions on how to run these unit tests.

cbiesinger: Added a comment at the top of the file. (It's just "python JsDbg_test.py")

PR #166, opened 2019-07-16
[gdb] Add an integration test

cbiesinger:
This tests that GDB successfully loads the python script and that its
various functions work correctly on a test program.

Uses the "dejagnu" tool, which has poor documentation but
seems otherwise well suited to testing command-line interfaces
like this one (gdb uses it for its tests too).

### DIFF
--- a/server/JsDbg.Gdb/testsuite/README
+++ b/server/JsDbg.Gdb/testsuite/README
@@ -1,0 +1,11 @@
+To run these tests, first install dejagnu: https://www.gnu.org/software/dejagnu/
+You can install it using "apt-get install dejagnu" or "yum install dejagnu".
+
+Then run "runtest" in this directory.
+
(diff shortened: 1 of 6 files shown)


cbiesinger:
This does find a bug in GetSymbolsForStackFrame which I will fix separately.

Also WriteMemoryBytes does not work if GDB was compiled with Python 2.

server/JsDbg.Gdb/testsuite/README
```diff
@@ -0,0 +1,5 @@
+To run these tests, install dejagnu and run "runtest" in this directory.
```
sanketj: Could you add some instructions on how to install dejagnu and point to some of its documentation?

cbiesinger: Done

cbiesinger:
Done

PR #170, opened 2019-07-21
[gdb] Move SFoo type definitions to a separate file

cbiesinger:
This allows reusing them for a future LLDB implementation


cbiesinger:
@speednoisemovement fyi

PR #172, opened 2019-07-25
[gdb] Look for dotnet in the path by default

cbiesinger:
This is probably more likely to work, especially if users install dotnet
using the package manager.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -28,8 +28,8 @@ Once you have cloned the JsDbg repository, please follow the steps below to buil
 
 #### Linux
 
-1. Get the [x64 .NET Core SDK binary](https://dotnet.microsoft.com/download) (click the x64 link in the left column in the table at the bottom).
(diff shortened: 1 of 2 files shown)


docs/CONTRIBUTING.md
```diff
@@ -28,8 +28,8 @@ Once you have cloned the JsDbg repository, please follow the steps below to buil
 
 #### Linux
 
-1. Get the [x64 .NET Core SDK binary](https://dotnet.microsoft.com/download) (click the x64 link in the left column in the table at the bottom).
-2. To build the server, run `make`. If you extracted `dotnet` into a different directory than `~/dotnet`, you need to instead run `make DOTNET=/path/to/your/dotnet/binary`. Alternatively, run `make install` (with DOTNET= as appropriate).
+1. Get the [x64 .NET Core SDK binary](https://dotnet.microsoft.com/download) (click the x64 link in the left column in the table at the bottom). If you don't install it as a package, make sure to add the directory to your $PATH (e.g. `export PATH=$PATH:$HOME/dotnet`)
```
sanketj: Maybe `export PATH=$PATH:/path/to/your/dotnet/binary` would be better? Up to you though.

cbiesinger: Hm... I think I like mine better because it emphasizes that you should use $HOME instead of ~

PR #176, opened 2019-07-27
[gdb] Add a GDB parameter to control verbose output

cbiesinger:
This is an easier-to-use way to set verbosity.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -96,7 +96,7 @@ import importlib
 importlib.reload(JsDbg)
 ```
 
-4. For debugging JsDbg itself, you may want to set verbose to True, either by editing `JsDbg.py` or running `python JsDbg.jsdbg.verbose = True` at the gdb prompt.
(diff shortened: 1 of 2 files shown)


docs/CONTRIBUTING.md
```diff
@@ -96,7 +96,7 @@ import importlib
 importlib.reload(JsDbg)
 ```
 
-4. For debugging JsDbg itself, you may want to set verbose to True, either by editing `JsDbg.py` or running `python JsDbg.jsdbg.verbose = True` at the gdb prompt.
+4. For debugging JsDbg itself, you may want to `set jsdbg-verbose on` at the GDB prompt. (Due to a GDB bug, you probably want to combine this with `set pagination off`).
```
sanketj: What's the GDB bug you are mentioning and how does turning off pagination help?

cbiesinger: I've linked the bug now. The issue is that verbose output produces a lot of output, triggering pagination. But GDB (well, libreadline) crashes when you trigger pagination from inside of an event posted by gdb.post_event, which is what we use.

sanketj: Ok, great. LGTM. Thanks.

PR #177, opened 2019-07-29
[gdb] Fix JsDbg_test

cbiesinger:
I forgot to run the test when I added the GDB parameter :(
(in commit a39c4908adf5fcdb6d1c334b08d8b7eebbdf5326)


cbiesinger:
Thanks, yes, it would be great to automatically run these.

PR #179, opened 2019-07-30
[gdb] Move non-GDB specific functionality into its own class

cbiesinger:


server/JsDbg.Gdb/JsDbgBase.py
```diff
@@ -0,0 +1,138 @@
+# This is the non-debugger specific base functionality of JsDbg.
```
sanketj: For this to be used by both GDB and LLDB, I think we have to move this out of JsDbg.GDB and into a location that both the GDB and LLDB servers can refer to. We do something similar to share core functionality between JsDbg.WinDbg and JsDbg.VisualStudio.

cbiesinger: Do you mind if I do that in a separate patch? Because I really should move the webserver code as well, and that would make this a fairly big patch.

For the other question -- this is a future-looking change

cbiesinger: Actually I'll just do a second commit on this PR. Will let you know when I'm done.

cbiesinger: OK, please take another look

server/JsDbg.Stdio/JsDbgBase.py
```diff
@@ -0,0 +1,140 @@
+# This is the non-debugger specific base functionality of JsDbg.
+# For use by GDB (or LLDB)-specific modules.
+
+import os.path
+import re
+import subprocess
+import threading
+
+class JsDbg:
+    """The debugger-independent functionality of JsDbg.
```
sanketj: nit: This comment is redundant with line 1. Consider removing one of them.

server/JsDbg.Stdio/JsDbgBase.py
```diff
@@ -0,0 +1,140 @@
...
+
+    It is designed to be simple to use; users only have to provide three things:
+    - A module instance that contains the functions that the webserver calls
+      (primarily DebuggerQuery)
+    - A post_event function that takes a callable class to execute on the
+      main thread.
+    - A function that gets called when the server crashes/exists, so any
+      global state can be cleaned up and the user alerted.
+
+    When JsDbg is  instantiated, it will look for the JsDbg webserver binary
```
sanketj: nit: extra space between "is" and "instantiated"

server/JsDbg.Stdio/JsDbgBase.py
```diff
@@ -0,0 +1,140 @@
...
+
+import os.path
+import re
+import subprocess
+import threading
+
+class JsDbg:
+    """The debugger-independent functionality of JsDbg.
+
+    It is designed to be simple to use; users only have to provide three things:
```
sanketj: Love it! :)

server/JsDbg.Stdio/Program.cs
```diff
@@ -16,8 +16,8 @@ class Program
 
             string extensionsDirectory = args[0];
 
-            // Inversion of control: Assume that this process has been started by python, stdio talks back to python and can ask it to do things.
-            GdbDebugger debugger = new GdbDebugger();
+            // This process has been started by python, stdio talks back to python and can ask it to do things.
```
sanketj: I know this comment has been around for a while, but can we update it to be more specific than "to do things"?

PR #182, opened 2019-08-02
[gdb] Better module handling

cbiesinger:
This makes GDB's module handling behave more similar to the one in
Windows:
- With all GDB versions, we check if a module is loaded before looking
  it up in GDB.
- In GDB 8.4+ (not yet released), we only look for a type/symbol in the
  requested module (this is not possible in older GDB versions).

### DIFF
--- a/server/JsDbg.Gdb/JsDbg.py
+++ b/server/JsDbg.Gdb/JsDbg.py
@@ -60,6 +60,42 @@ class GdbNamedSymbol(JsDbgTypes.SNamedSymbol):
             GdbSymbolResult(symbol, frame))
 
 
+def FindObjfileForName(name):
+    for objfile in gdb.objfiles():
+        if JsDbgBase.FormatModule(objfile.filename) == name:
+            return objfile
+    return None
+
+
+def FindGdbSymbol(module, symbol):
+    objfile = FindObjfileForName(module)
+    if objfile is None:
+        return None
+    # GDB 8.4 and later let us look up symbols per-objfile; use that if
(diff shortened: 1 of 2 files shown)


server/JsDbg.Gdb/JsDbg.py
```diff
@@ -200,21 +243,18 @@ def LookupField(module, type, field):
         fields = [f for m in match for f in m.type.fields()]
 
 def LookupGlobalSymbol(module, symbol):
-    # We can't use lookup_global_symbol because that does not work for symbols
-    # with local linkage, such as those in an anonymous namespace.
-    # https://sourceware.org/bugzilla/show_bug.cgi?id=24474
-    (sym, _) = gdb.lookup_symbol(symbol)
+    sym = FindGdbSymbol(module, symbol)
```
sanketj: Does this get slower now that we incur the additional perf cost of gdb.objfiles()?

sanketj: Same for the calls to FindGdbType above/below.

cbiesinger: gdb.objfiles is pretty quick. I think this should be a net win because we now no longer call lookup_symbol when the objfile isn't loaded, and especially on GDB 8.4 (current Git, or whenever it will be released), where we only search one objfile instead of all of them, which is the slow part.

sanketj: Ok. SGTM.

PR #184, opened 2019-08-02
[gdb] Don't try to set verbose on a non-running jsdbg

cbiesinger:


server/JsDbg.Gdb/JsDbg.py
```diff
@@ -407,7 +407,10 @@ def __init__(self):
 
     def get_set_string(self):
         global jsdbg
-        jsdbg.verbose = self.value
+        if jsdbg is not None:
+            # If JsDbg is already running, update is verbosity.
```
sanketj: nit: update **it's** verbosity

cbiesinger: Updated to "its"

sanketj: Yes, sorry. Should be its. :)

PR #190, opened 2019-08-02
[gdb] Put the resulting deb in the current directory

cbiesinger:
Unfortunately this will still put a .changes file in the parent
directory and I don't know how to make it not do that.

Also renames the resulting deb to "jsdbg-gdb.deb" (and ensures that
the package name is changed to that)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,10 @@ all:
 
 deb: clean
 	@echo '*** See debian/README.source for more information and help with troubleshoting. ***'
-	dpkg-buildpackage -us -uc
+	@# We can't use --filename=jsdbg-gdb.deb in BUILDPACKAGE_OPTS because too many steps
+	@# depend on keeping the filename as-is. So we'll just rename it afterwards.
+	BUILDPACKAGE_OPTS="--destdir=." dpkg-buildpackage -us -uc --build=binary --buildinfo-option=-u. --changes-option=-u.
+	mv jsdbg-gdb_*.deb jsdbg-gdb.deb
 
 %:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -3,7 +3,10 @@ all:
 
 deb: clean
 	@echo '*** See debian/README.source for more information and help with troubleshoting. ***'
-	dpkg-buildpackage -us -uc
+	@# We can't use --filename=jsdbg-gdb.deb in BUILDPACKAGE_OPTS because too many steps
+	@# depend on keeping the filename as-is. So we'll just rename it afterwards.
+	BUILDPACKAGE_OPTS="--destdir=." dpkg-buildpackage -us -uc --build=binary --buildinfo-option=-u. --changes-option=-u.
+	mv jsdbg-gdb_*.deb jsdbg-gdb.deb
```
sanketj: Should this be "mv **jsdbg_*.deb** jsdbg-gdb.deb"? The name of .deb package produced currently doesn't seem to include "gdb".

sanketj: Or is that what the package name change in the control file is about?

cbiesinger: The change to debian/control makes it produce jsdbg-gdb_0.9.0_amd64.deb

sanketj: Ok, makes sense.

cbiesinger: Technically no, but it is necessary to make jsdbg show up as jsdbg-gdb in the list of installed packages (e.g. dpkg -l), which I thought would be good to change if we rename the .deb file like that.

sanketj: Ok. Sounds good.

PR #195, opened 2019-08-19
[gdb] Better handle typedefs to anonymous unions

cbiesinger:
When compiled with clang, for a typedef to an anonymous union, gdb
will not be able to find a name and so 'name' is None. When that
happens, we should just not call strip_typedefs to keep a name
to refer to the type.

Because this only happens with clang, I have not written a testcase
for this bug. I did file a clang bug (see link in comment).

This shows up with base::internal::LockImpl::NativeHandle.

### DIFF
--- a/server/JsDbg.Gdb/JsDbg.py
+++ b/server/JsDbg.Gdb/JsDbg.py
@@ -113,7 +113,17 @@ def IsFunctionPointer(t):
 
 
 def FormatType(symbol_type):
-    t = symbol_type.strip_typedefs()
+    # The type may be a typedef to an anonymous union (e.g.
+    # base::internal::LockImpl::NativeHandle). In that case, keep the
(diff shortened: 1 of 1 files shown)


server/JsDbg.Gdb/JsDbg.py
```diff
@@ -113,7 +113,17 @@ def IsFunctionPointer(t):
 
 
 def FormatType(symbol_type):
-    t = symbol_type.strip_typedefs()
+    # The type may be a typedef to an anonymous union (e.g.
+    # base::internal::LockImpl::NativeHandle). In that case, keep the
+    # typedef name so we have *some* name to refer to this type.
+    # (This is only and issue with clang,
```
sanketj: nit: only *an* issue

cbiesinger: Done

PR #202, opened 2019-11-05
Support older GDB versions

cbiesinger:
Unfortunately, thread/process change events won't work, but it's
probably better than throwing an exception.


server/JsDbg.Gdb/JsDbg.py
```diff
@@ -456,7 +456,9 @@ def PromptHandler():
     gdb.events.stop.connect(StoppedHandler)
     gdb.events.cont.connect(ContHandler)
     gdb.events.exited.connect(ExitHandler)
-    gdb.events.before_prompt.connect(PromptHandler)
+    # before_prompt was added in GDB 8.0.
```
sanketj: nit: might be clearer to say: before_prompt is only supported on GDB 8.0 and above

cbiesinger: Thanks, fixed.

PR #205, opened 2020-06-18
Support global lookup of symbols within scopes (namespaces, outer classes, etc.) on Windows

sanketj:
The `DbgObject.global` API takes a scope parameter; however, it is currently ignored on Windows. With the latest compiler changes affecting Chromium symbols like `g_frame_map`, the scope now needs to be considered when querying for symbols on Windows. To that end, this change adjusts the symbol name to include the scope prefix before querying the debugger/symbol files on Windows.

There is an additional complication that needs to be addressed to handle anonymous namespaces correctly. On Linux, an anonymous namespace is annotated as `(anonymous namespace)`, whereas on Windows, it is annotated as `` `anonymous namespace'``. To handle this discrepancy, the `DbgObject.global` API now takes an array of scope strings (ordered [outermost scope, ..., innermost scope]), and within this array, an anonymous namespace is simply written as "anonymous namespace". Then, this `scopes` array will be transformed into a scope prefix string, with the Linux/Windows specific code resolving anonymous namespaces in platform-specific ways.

### DIFF
--- a/extensions/dbgobject/core/dbgobject.js
+++ b/extensions/dbgobject/core/dbgobject.js
@@ -122,7 +122,7 @@ Loader.OnLoad(function() {
             {name:"moduleName", type:"string", description:"The name of the module containing the symbol."},
             {name:"symbol", type:"string", description:"The global symbol to lookup."},
             {name: "typeName", type:"string", description: "(optional) The type name of the symbol to look up."},
-            {name: "scopes", type:"array of strings", description: "(optional) Array of scopes of the symbol to look up (namespace, outer class, etc.). Order: [outermost scope, ..., innermost scope]."}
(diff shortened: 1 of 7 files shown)


extensions/dbgobject/core/dbgobject.js
```diff
@@ -122,12 +122,12 @@ Loader.OnLoad(function() {
             {name:"moduleName", type:"string", description:"The name of the module containing the symbol."},
             {name:"symbol", type:"string", description:"The global symbol to lookup."},
             {name: "typeName", type:"string", description: "(optional) The type name of the symbol to look up."},
-            {name: "scope", type:"string", description: "(optional) The scope of the symbol to look up (namespace, outer class, etc.). Required on some platforms."}
+            {name: "scopes", type:"array of strings", description: "(optional) Array of scopes of the symbol to look up (namespace, outer class, etc.). Order: [outermost scope, ..., innermost scope]."}
```
cbiesinger, line 125: Should this mention the special treatment of "anonymous namespace"?

sanketj: Good call! Will update this to mention how anonymous namespaces should be declared.

PR #207, opened 2021-04-02
Modernize debian packaging

cbiesinger:
- Depend on the .NET SDK 5.0 instead of 2.2
- Run the Python 2 tests using `python2` instead of just `python`


sanketj (approved):
LGTM

PR #209, opened 2021-08-24
Make debian build process work with current dotnet releases.

cbiesinger:


debian/rules
```diff
@@ -5,7 +5,10 @@ export DH_VERBOSE = 1
 
 export DOTNET_CLI_TELEMETRY_OPTOUT = 1
 export DOTNET_CLI_HOME := $(shell pwd)
-export RESTOREFLAGS := -s /usr/share/dotnet/sdk/NuGetFallbackFolder
+# This does not work anymore in newer dotnet releases.
+# We previously used it to avoid making network connections during
+# package build.
+# export RESTOREFLAGS := -s /usr/share/dotnet/sdk/NuGetFallbackFolder
```
sanketj, line 11: Is anyone still using debian with older dotnet versions?

cbiesinger: Probably not (can't install them on current debian because they depend on an ICU package version that has been removed from debian). Are you implying I should just remove those lines?

sanketj: I was trying to say two things: (1) if someone could still be using an older version, then they might be broken by this change, (2) and yes, if indeed no one needs this, then we should just delete it.

cbiesinger: Makes sense. Done. Older versions won't be broken by the change either -- the build will just require a network connection.

sanketj: Thanks, LGTM.

sanketj (approved):

